PR #27, opened 2021-05-08
Update LibMobileCoin to v1.1.0-pre2

kylefleming:


sugargoat (approved):

Example/Podfile
```diff
@@ -26,7 +26,7 @@ target 'Example' do
   # pod 'MobileCoin', git: 'https://github.com/mobilecoinofficial/MobileCoin-Swift.git'
   # pod 'MobileCoin/Core', git: 'https://github.com/mobilecoinofficial/MobileCoin-Swift.git', testspecs: ['Tests', 'IntegrationTests']
 
-  pod 'LibMobileCoin', binary: true
+  pod 'LibMobileCoin', '~> 1.0-pre', binary: true
```
jcape, line 29: This was the thing I was asking about, should this be 1.1-pre?

Example/Podfile.lock
```diff
@@ -104,7 +104,7 @@ PODS:
 DEPENDENCIES:
   - gRPC-Swift
   - Keys (from `Pods/CocoaPodsKeys`)
-  - LibMobileCoin
+  - LibMobileCoin (~> 1.0-pre)
```
jcape, line 107: Same q.

jcape (approved):
Approved conditional on those two "1.0-pre" versions being right... :-)

Example/Podfile
kylefleming: No, the example works with any version. CocoaPods requires specifying a version with a pre-release suffix in order to select a pre-release, so instead of `>0` or `>1`, I set it to `~> 1.0-pre`, which means any pre-release version greater than 1, but less than 2.

jcape: Ah, OK

PR #29, opened 2021-05-11
[SWIFT-167] Implement use of dynamic fees

kylefleming:
This PR adds use of the BlockchainAPI service, and uses the new `LastBlockInfoResponse.minimumFee` to determine the minimum fee to use. It defaults to the old hard-coded minimum fee if `minimumFee` is missing from the `LastBlockInfoResponse` response, allowing this SDK version to be compatible with both the old server/enclave deployment and the new one with lower fees.

A future PR will include caching of the reported fee.


sugargoat (approved):

voloshyn (approved):
LGTM

PR #64, opened 2021-09-03
[Draft] Network Robustness 

the-real-adammork:
Builds on changes originally from Kyle and then Alex.

- Add an `.http` option so integrating apps can pass in an object conforming to `HttpRequester` and we will use it for all network processing
- Includes changes for new `fog` (now `mobilecoin`) in regards to the `CompressedCommitment` protobuf removal. 

Related Submodule PRs:

- https://github.com/mobilecoinofficial/libmobilecoin-ios-artifacts/pull/1
- https://github.com/mobilecoinfoundation/mobilecoin/pull/893


the-real-adammork:
Screen Shot 2021-09-03 at 4 46 54 PM

PR #84, opened 2021-11-03
Add seperate HTTP & GRPC paths to integration tests. Adjust init and …

the-real-adammork:
Soundtrack of this PR: [Roy Ayers - Searching](https://www.youtube.com/watch?v=R8FSBbRNhEk)

### Motivation

After adding network robustness (HTTP) support, the integration tests still only one `TransportProtocol` option which would be its default value as set in the `NetworkConfig.swift`

### In this PR

Add separate test functions for each protocol option. Keep existing test functions but parameterize them to accept a `TransportProtocol` and then add new test functions for each type - GRPC & HTTP.

### Future Work
- Possibly split the integration tests into separate targets/groups for GRPC & HTTP ? This might be required for splitting GRPC into its own build-target as has been requested by Signal.

### DIFF
--- a/Sources/MobileCoinClient.swift
+++ b/Sources/MobileCoinClient.swift
@@ -307,7 +307,8 @@ extension MobileCoinClient {
             fogViewAttestation: Attestation,
             fogKeyImageAttestation: Attestation,
             fogMerkleProofAttestation: Attestation,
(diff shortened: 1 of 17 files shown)


voloshyn (approved):
LGTM

PR #93, opened 2021-12-17
Create an `HTTP` only target for smaller library size

the-real-adammork:
Soundtrack of this PR: [Busta 929 - Mmapula (feat. Mzu M)](https://www.youtube.com/watch?v=rgICb0vEQMk)

### Motivation

Some SDK consumers would like a smaller library size with the trade-off of removing GRPC networking.

App Thinning Report Comparison:

| Example Project      | Protocols | Uncompressed   |   Compressed  |
| :---        |    :---    |      :---    |   :---  |
| **ExampleHTTP**   | **HTTP**      |  **5.7mb**   |  **20.1mb** |
| Example      | GRPC + HTTP       |  10.0mb  |   37.4mb |

### In this PR
- Obfuscate GRPC specific dependencies behind protocols
- Add HTTP & GRPC connection factories that can return "Empty" connections if the protocol is not available.
- Update integration unit tests to test on supported protocols
- Add protocol specific unit tests
- LibMobileCoin updated as well https://github.com/mobilecoinofficial/libmobilecoin-ios-artifacts/pull/6

### Future Work
- Fix/Add GRPC only target


the-real-adammork:
@voloshyn tests passing, updated to include the new `ExampleHTTP` project which is configured to use the "HTTP Only" targets.

voloshyn (approved):
LGTM!

the-real-adammork:
`pod lib lint` is passing, merging and then pushing to cocoapods

PR #98, opened 2022-01-10
Use same ssh key for publishing docs

the-real-adammork:
### Motivation

Publish needs an SSH key with write access.

### In this PR
- Specify the SSH key already in the project for publishing docs


MCrank (approved):
LGTM

PR #99, opened 2022-01-11
Add unit tests for TransferPayload and PaymentRequest Printable encodings

carybakker:
* TransferPayload <-> Printable_TransferPayload
 * PaymentRequest <-> Printable_PaymentRequest

### Motivation

Increase unit test code coverage

### In this PR
* Added TransferPayloadTests and PaymentRequestTests to UnitTests

### DIFF
--- a/Tests/Unit/Encodings/PaymentRequestTests.swift
+++ b/Tests/Unit/Encodings/PaymentRequestTests.swift
@@ -160,4 +160,16 @@ class PaymentRequestTests: XCTestCase {
         }
     }
 
+    func testDecodingFromPrintableWithInvalidPublicAddress() throws {
+        // create Printable_PaymentRequest
+        var printablePaymentRequest = Printable_PaymentRequest()
+        printablePaymentRequest.publicAddress = External_PublicAddress()
+
+        // create PaymentRequest from Printable_PaymentRequest
+        let paymentRequest = PaymentRequest(printablePaymentRequest)
+
+        // Validate
+        XCTAssertNil(paymentRequest)
+    }
+
 }
--- a/Tests/Unit/Encodings/PaymentRequestTests.swift
+++ b/Tests/Unit/Encodings/PaymentRequestTests.swift
@@ -8,160 +8,158 @@ import XCTest
 
 class PaymentRequestTests: XCTestCase {
 
-    static let publicAddress = (try? PublicAddress.Fixtures.Init())!.accountKey.publicAddress
-    static let externalPublicAddress = External_PublicAddress(publicAddress)
     static let value: UInt64 = 123
     static let memo = "test memo"
 
     func testEncodingToPrintableNoValueNoMemom() throws {
-        XCTAssertNoThrow(evaluating: {
-            // create PaymentRequest
-            let paymentRequest = PaymentRequest(publicAddress: Self.publicAddress)
-            XCTAssertNotNil(paymentRequest)
+        let publicAddress = try PublicAddress.Fixtures.Init().accountKey.publicAddress
 
-            // create Printable_PaymentRequest from PaymentRequest
-            let printablePaymentRequest = Printable_PaymentRequest(paymentRequest)
+        // create PaymentRequest
+        let paymentRequest = PaymentRequest(publicAddress: publicAddress)
+        XCTAssertNotNil(paymentRequest)
 
-            // Validate
-            XCTAssertNotNil(printablePaymentRequest)
-            XCTAssertEqual(printablePaymentRequest.publicAddress, Self.externalPublicAddress)
-            XCTAssertEqual(printablePaymentRequest.value, 0)
-            XCTAssertEqual(printablePaymentRequest.memo.count, 0)
-        })
+        // create Printable_PaymentRequest from PaymentRequest
+        let printablePaymentRequest = Printable_PaymentRequest(paymentRequest)
+
+        // Validate
+        XCTAssertNotNil(printablePaymentRequest)
+        XCTAssertEqual(printablePaymentRequest.publicAddress, External_PublicAddress(publicAddress))
+        XCTAssertEqual(printablePaymentRequest.value, 0)
+        XCTAssertEqual(printablePaymentRequest.memo.count, 0)
     }
 
     func testDecodingFromPrintableNoValueNoMemo() throws {
-        XCTAssertNoThrow(evaluating: {
-            // create Printable_PaymentRequest
-            var printablePaymentRequest = Printable_PaymentRequest()
-            printablePaymentRequest.publicAddress = Self.externalPublicAddress
+        let publicAddress = try PublicAddress.Fixtures.Init().accountKey.publicAddress
 
-            // create PaymentRequest from Printable_PaymentRequest
-            let paymentRequest = PaymentRequest(printablePaymentRequest)
+        // create Printable_PaymentRequest
+        var printablePaymentRequest = Printable_PaymentRequest()
+        printablePaymentRequest.publicAddress = External_PublicAddress(publicAddress)
 
-            // Validate
-            XCTAssertNotNil(paymentRequest)
-            if let paymentRequest = paymentRequest {
-                XCTAssertEqual(paymentRequest.publicAddress, Self.publicAddress)
-                XCTAssertNil(paymentRequest.value)
-                XCTAssertNil(paymentRequest.memo)
-            }
-        })
+        // create PaymentRequest from Printable_PaymentRequest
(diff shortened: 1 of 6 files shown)


Tests/Unit/Encodings/PaymentRequestTests.swift
```diff
@@ -0,0 +1,167 @@
...
+
+class PaymentRequestTests: XCTestCase {
+
+    static let publicAddress = (try? PublicAddress.Fixtures.Init())!.accountKey.publicAddress
+    static let externalPublicAddress = External_PublicAddress(publicAddress)
+    static let value: UInt64 = 123
+    static let memo = "test memo"
+
+    func testEncodingToPrintableNoValueNoMemom() throws {
+        XCTAssertNoThrow(evaluating: {
```
the-real-adammork: I think the `XCTAssertNoThrow` is unnecessary. If any line in the test throws it will be considered a failure by XCTest.

Tests/Unit/Encodings/PaymentRequestTests.swift
```diff
@@ -0,0 +1,167 @@
...
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import LibMobileCoin
+@testable import MobileCoin
+import XCTest
+
+class PaymentRequestTests: XCTestCase {
+
+    static let publicAddress = (try? PublicAddress.Fixtures.Init())!.accountKey.publicAddress
```
the-real-adammork: Lets move this in-line to each test where its needed. You can change it to something like:

`let publicAddress = try PublicAddress.Fixtures.Init().accountKey.publicAddress`

The force-cast can go away, and if it fails for some reason it will show up in the test results.

Tests/Unit/Encodings/TransferPayloadTests.swift
```diff
@@ -0,0 +1,191 @@
...
+//
+
+import LibMobileCoin
+@testable import MobileCoin
+import XCTest
+
+class TransferPayloadTests: XCTestCase {
+
+    // The base64Encoded string below will decode to 32 bytes of data for a valid Data32
+    static var entropyData = Data(base64Encoded: "ajaEQTHHDeZEZDk1rGYQRF0ErcpmcPa7buRpNchz4hQ=")!
```
the-real-adammork: we should avoid force-casts when possible.

the-real-adammork (changes requested):
see comments

Tests/Unit/Encodings/PaymentRequestTests.swift
carybakker: switched to declaring locally in-line

Tests/Unit/Encodings/TransferPayloadTests.swift
carybakker: Ya, this was annoying me a bit too....

I didn't want to put these (e.g. also the ristrettoCompressed) values into the local functions - not sure how costly, say, the External_Compressed(..) call is...   Also they're used in every single test.... 

So, I wanted to factor them out to static so they're only intialized once for the entire set of tests....

The problem is there's no constructor for the XCTestCase that we can use to initialize them, yet they do need to be initialized.

I had started off with using var instead of let (this one was left var by accident), and used an initial empty Data() instance, that I then overwrote with the Class level setUp function... 

So this, while not ideal, seemed to be the lesser of the two evils.

So it's kind of like we have these 3 options:
 1. declare the string constants as static for the class, and create the entropy and ristretto values in each test (and duplicate the code, which bugs me too b/c it's not DRY)
 1. change from let to var, meaning the values could theoretically be changed by the individual tests (which I don't like), and then initialize with a dummy Data() that's overwritten in the class level setUp function, but that also feels a bit janky
 1. add the forced unwrap for the optional, which given the known hardcoded data is known to be correct, seemed not too bad.

What are your thoughts based on the above?

the-real-adammork: The existing pattern is to define these things in a "Fixtures" file, then reference them in the tests:

```
  Tests/Common/Fixtures/Network/Connection/TestBlockchainService.swift
  Tests/Common/Fixtures/Network/Attestation/Attestation+Fixtures.swift
  Tests/Common/Fixtures/Network/Attestation/AttestAke+Fixtures.swift
  Tests/Common/Fixtures/Crypto/VersionedCryptoBox+Fixtures.swift
  Tests/Common/Fixtures/Transaction/Transaction+Fixtures.swift
  Tests/Http/FixturesData/ConnectionSession+HttpFixtures.swift
  Tests/Grpc/FixturesData/ConnectionSession+GrpcFixtures.swift
  Tests/Common/Fixtures/Fog/Report/FogResolver+Fixtures.swift
  Tests/Common/Fixtures/Account/PublicAddress+Fixtures.swift
  Tests/Common/Fixtures/Network/NetworkConfig+Fixtures.swift
  Tests/Common/Fixtures/Transaction/Receipt+Fixtures.swift
  Tests/Common/Fixtures/Fog/View/FogRngKey+Fixtures.swift
  Tests/Common/Fixtures/Account/AccountKey+Fixtures.swift
  Tests/Common/Fixtures/MobileCoinClient+Fixtures.swift
  Tests/Common/Fixtures/Fog/View/FogRng+Fixtures.swift
  Tests/Common/Fixtures/Account/Account+Fixtures.swift
  Tests/Common/Fixtures/Network/NetworkPreset.swift
  Tests/Common/Fixtures/Ledger/TxOut+Fixtures.swift
  Tests/Common/Fixtures/Encodings/Base58Coder.swift
  Tests/Common/Fixtures/Encodings/MobUri.swift
```

Its a decent way to get all the static test data into one place so it can be re-used.

Tests/Unit/Encodings/TransferPayloadTests.swift
```diff
@@ -0,0 +1,176 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import LibMobileCoin
+@testable import MobileCoin
+import XCTest
+
+class TransferPayloadTests: XCTestCase {
+    static var
```
the-real-adammork:
```suggestion
```

Tests/Unit/Encodings/TransferPayloadTests.swift
```diff
@@ -0,0 +1,176 @@
...
+    
+    // The base64Encoded string below will decode to 32 bytes of data for a valid Data32
+    static var entropyData: Data(base64Encoded: "ajaEQTHHDeZEZDk1rGYQRF0ErcpmcPa7buRpNchz4hQ=")!
```
the-real-adammork:
```suggestion
    static var entropyData = Data(base64Encoded: "ajaEQTHHDeZEZDk1rGYQRF0ErcpmcPa7buRpNchz4hQ=")!
```

Tests/Unit/Encodings/TransferPayloadTests.swift
```diff
@@ -0,0 +1,176 @@
...
+    static let entropyData32: Data32 = Data32(entropyData)!
```
the-real-adammork:
```suggestion
    static let entropyData32 = Data32(entropyData)!
```

the-real-adammork:
Can we write a failing test for `TransferPayload` that has both "bip39" and "rootEntropy" set ?

Tests/Unit/Encodings/TransferPayloadTests.swift
carybakker: Awesome - switched to using Fixtures for the test data.

Also, if the test data creation fails, I throw an exception with specific context around what, exactly, failed, and this info will then be logged to the test output so it will be easy to debug.

e.g., a dev accidentally deletes a character from the base64 encoded strings used to generate a test value, so the test value is nil....

I didn't want to add checks to also verify the test data within the tests... but the values *could* be nil, and that may still allow the tests to pass (b/c the verified values will still be equal). 

This way, if something changes or is wrong with the input, the tests will fail, and the specific problem will be reported to save time debugging.

Tests/Unit/Encodings/TransferPayloadTests.swift
carybakker: meh, this was an accidental commit - sorry about that.

Tests/Unit/Encodings/TransferPayloadTests.swift
carybakker: fixed with new Fixture impl

carybakker:
> Can we write a failing test for `TransferPayload` that has both "bip39" and "rootEntropy" set ?

the-real-adammork (approved):
Looks good !

PR #104, opened 2022-01-27
Network Robustness Improvements & HTTPS Cert-pinning reference implementation

the-real-adammork:
Soundtrack of this PR: [Girl Unit - Wut (Claude Vonstroke Remix) (Butt Naked Mix)](https://www.youtube.com/watch?v=_CRQ888T2Zk)

### Motivation

Some SDK changes were found to be necessary for real-world integration of the new `CoreHTTP` (http-only) build target.  

Additionally the HTTPS implementation of SSL certificates (for cert-pinning) was lacking. This PR addresses the lack of a dependency free "reference implementation" of cert-pinning for our "default HTTPS requester". 

### In this PR
- Add required initializer to `HTTPResponse` for real-world integration
- Add dependency-free certificate pinning reference implementation based on Apple's Security Framework
- Rename SSLCertificate protocols to encompass the range of conforming entities
- Re-enable HTTPS in cert-pinning unit tests
- Store `SSLCertificates` for both transport protocol options
- Expand HttpRequester protocol with optional functions for setting fog/consensus trust roots
- Set trustRoots on HttpRequester every time either of them changes 
- Add `make lock` to quickly re-generate example projects & lock files
- Increase timeouts on some unit tests


voloshyn (approved):
LGTM

voloshyn:
@dolanbernard I've added the ticket for Android to add a default implementation of the HttpRequester too

Sources/Network/Certificates+Extensions.swift
```diff
@@ -67,3 +67,39 @@ extension SecCertificate {
...
     }
 }
+
+extension SecTrust {
+    public var certificateCount : Int {
+        SecTrustGetCertificateCount(self)
+    }
+
+    public var certificateTrustChain : [SecCertificate] {
+        [Int](0..<certificateCount).compactMap({
```
carybakker: minor nit, no need for parens to wrap the transform block

Sources/Network/Certificates+Extensions.swift
```diff
@@ -0,0 +1,105 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+extension Data {
+    public func asPinnedCertificate() -> Result<SecCertificate, InvalidInputError> {
```
carybakker: Just minor naming feedback here - 

These functions are general cert functions, but the fact that a cert is a pinned cert really just meta-info about how we're using the cert...

Nothing about the fact that the cert is being used as a pinned cert matters here, right?  Can we drop the 'pinned' from these function names as the functions may be useful with certs for some reason outside of the context of pinning? (likely the only reason we'll use them, but still :) )

the-real-adammork: makes sense, Ill change to asSecCertificate and pinnedKeys -> secKeys/asSecKeys

Sources/Network/Certificates+Extensions.swift
the-real-adammork: fixed

Sources/Network/HTTPS/DefaultHttpRequester.swift
```diff
@@ -0,0 +1,148 @@
...
+//
+
+import Foundation
+
+public class DefaultHttpRequester: NSObject, HttpRequester {
+    private var fogTrustRoots: SecSSLCertificates?
+    private var consensusTrustRoots: SecSSLCertificates?
+    
+    private var pinnedKeys: [SecKey] {
+        [fogTrustRoots, consensusTrustRoots].compactMap {
```
carybakker, line 12: For the pinned certs, seems like we're commingling here and just combining the list of pinned certs into a single list that's used across all requests.

Seems like to do pinning more securely, we should differentiate between fog requests and consensus requests and validate with the respective root certs...

May not really be a big deal - but... if we separate them then it would take more than access to a single server's cert to compromise the traffic for both fog and consensus...

And if it's ok to actually combine them, then why do we have two different trust roots instead of sharing one set of trusted roots across all servers?

carybakker (commented):
Still going over things, but just sharing what I have for now

PR #105, opened 2022-01-27
Initial commit of addition of url load balancing (for discussion not merge yet):

carybakker:
* currently using fixed RandomUrlLoadBalancer instance, will update
   to use load balancing protocol (thinking that over now)

 * currently using several forced unwraps of optionals to avoid complicating
   code with a lot of error checking and nil handling of optionals, introduced
   by the use of the array of urls (that may be empty) as well as the use
   of calls that may throw that should not... thinking over how best to structure
   these changes to avoid a lot of unnecessary error checking, however, also thinking
   that the load balancer can conceivably end up without any URL options after
   significant failures?

 * still need to add in the getNext() calls on erros when making calls to the URLs

### Motivation

< The motivation for the changes in this PR. "Currently we...", "This is needed because..." >

### In this PR
* < Additions, removals, fixes, features >

< Ticket status, e.g. "fixes #issue number" >

### Future Work
* < Out of scope non-goals for this PR >
* < These should be links to tickets. If the tickets do not exist, make them. >


the-real-adammork:
is it possible to keep AttestedConnectionConfig protocol as-is but add new var/ref/init-param to pass the loadBalancer then change the implementation for the "urlTyped" to pull from the loadBalancer as a computed variable ? would be a lower impact change. So the init for AttestedConnctionConfig still changes as you have it but the protocol var urlTyped can stay and be computed from LoadBalancer

Sources/Network/RandomUrlLoadBalancer.swift
```diff
@@ -0,0 +1,29 @@
...
+    
+    init(urls: [Url]) throws {
+        guard !urls.isEmpty else {
+            throw InvalidInputError("urls array must not be empty")
+        }
+        self.urlsTyped = urls
+        _ = nextUrl()
+    }
+    
+    func nextUrl() -> Url? {
```
the-real-adammork: add `@discardableResult` to nextUrl() , will let you call and ignore return value so line 22 would be "nextUrl()"

carybakker: Ah, nice... thanks!

carybakker: And I see what you mean about keeping the protocol as-is...  good point to minimize the impact.  Will ping you on Slack to discuss the details.

carybakker:
Closing this initial PR, switching to new branch task/ios-load-balancer-squashed and new PR https://github.com/mobilecoinofficial/MobileCoin-Swift/pull/114

PR #109, opened 2022-02-04
Adjust Public API - Remove Error conformance from `HttpStatus`

the-real-adammork:
Soundtrack of this PR: [Introbeats - Word](https://soundcloud.com/musicismysanctuary/exclusive-premiere-introbeatz-word)

### Motivation

`HttpStatus` conforms to the `Error` protocol even though it's not neccessary.  This adds extra complexity for SDK consumers who want to wrap all publicly exposed `Error`s. 

### In this PR

Remove `Error` conformance from `HttpStatus`, previously early exit errors (before the request was sent) were being represented by HttpStatus == 1. This was a hack so those early exit errors are now stored on the `HttpCallResult` as `error`. Code paths that return early exit errors have "assertionFailure" logging so they can be identified in debug.

Code paths that process `HttpCallResult` now check for a `.some(HttpStatus)` before continuing. A nil value is assumed to be an early exit error and is handled as a generic server error for now. 

### Future Work
* Using `Result<Success, Failure>` in place of an `HttpCallResult` might be better because the early exit paths can return `InvalidInputError` instead of `HttpCallResult`.  That change would be high impact because all of the Http networking code (including "generated" client files) would need to be updated. The motivation todo any further changes would be for better code semantics & readability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.0-pre9] - 2022-02-03
+
+### Changed
+
+- Simplified Publicly Exposed Error's conformers.
+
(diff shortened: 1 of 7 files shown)


voloshyn (approved):
LGTM

PR #110, opened 2022-02-04
Verify Correctness of Reconstructed Commitment, Update proto files

the-real-adammork:
Soundtrack of this PR: [Tilman - Just A Motion Away](https://soundcloud.com/moskalus/premiere-tilman-just-a-motion-away-shall-not-fade)

### Motivation

A previous change added a way to reconstruct the commitment data locally with info already sent in the TxOut protobuf. The comparison of its crc32 checksum was missing and needs to be added.

LibMobileCoin is updated with `mobilecoin` latest master which has some changes in the `consensus_common` proto files.

### In this PR
* Wrapper function around new function that computes the crc32 from a "CompressedCommitment"
* Add matching crc32's as a pre-condition to creating a `PartialTxOut`
* Update naming of `minimumFee` to `mobMinimumFee` (which is now deprecated) 

### Future Work
* Merge in changes in the `mobilecoin` repo, and the `libmobilecoin-ios-artifacts` repo
* Migrate minimum fee code to pull from the new map structure.


Sources/Ledger/PartialTxOut.swift
```diff
@@ -57,4 +58,21 @@ extension PartialTxOut {
...
+        let sentCommitment = txOutRecord.txOutAmountCommitmentData
+        let sentCrc32 = txOutRecord.txOutAmountCommitmentDataCrc32
+        let sentCommitmentCrc32: UInt32 = {
+            guard let sentCommitment32 = Data32(sentCommitment) else { return nil }
+            return TxOutUtils.calculateCrc32(from: sentCommitment32)
+        }() ?? .emptyCrc32
+        
+        let reconstructedCrc32 = TxOutUtils.calculateCrc32(from: reconstructed)
+        
+        return reconstructedCrc32 == sentCrc32 || reconstructedCrc32 == sentCommitmentCrc32
```
carybakker: Why do we use || instad of && here?

the-real-adammork: because the crc32 isn't always sent in the protobuf

if it is sent then we have to reconstruct the commitment data from the other fields in the TxOut (and our account keys) and then calculate the crc32 from the reconstructed commitment for comparison.

If crc32 is not sent (current + soon to be deprecated) then we still reconstruct the commitment data and then calculate crc32 for sent & reconstructed and compare.

Only one side can be true, which is sufficient.

voloshyn (approved):
Checking for crc32 presence is a correct implementation, but for the historical context: the crc32 was not present for the transition period from a full tx_out_amount_commitment_data to reconstructed, but it is always sent back now.
https://github.com/mobilecoinofficial/android-sdk/blob/842f9194a736c24ed2bc7ed9c4e8acc1c09c2310/android-sdk/src/main/proto/view.proto#L260

Sources/Ledger/TxOutUtils.swift
```diff
@@ -101,6 +99,35 @@ enum TxOutUtils {
...
+                    &errorPtr)
+            }) {
+            case .success:
+                return crc32
+            case .failure(let error):
+                switch error.errorCode {
+                case .invalidInput:
+                    // Safety: This condition indicates a programming error and can only
+                    // happen if arguments to mc_tx_out_commitment_crc32 are supplied incorrectly.
+                    logger.assertionFailure("error: \(redacting: error)")
```
the-real-adammork, line 120: @carybakker I changed these to assertionFailures which will act like "fatalError" during debug, but not in release.

Sources/Ledger/TxOutUtils.swift
```diff
@@ -101,6 +99,35 @@ enum TxOutUtils {
         }
     }
 
+    static func calculateCrc32(
```
the-real-adammork, line 102: Moved this call back to TxOutUtils so it fits in better with other similar calls

Sources/Utils/Data/Data+CommitmentCrc32.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+extension Data {
+    // this function is only valid in the following case:
+    //  - the Data instance is also a valid Data32
+    //  - the Data instance represents a CompressedCommitment
+    var commitmentCrc32: UInt32? {
```
the-real-adammork, line 11: made the extension name less verbose

Sources/Ledger/TxOutUtils.swift
carybakker: ah, k... sounds good

Sources/Ledger/TxOutUtils.swift
carybakker: @the-real-adammork I don't think anything is using this any longer - since you can just do <data>.commitmentCrc32 now, anyplace that was using this is now using the commitmentCrc32 ... or did I miss something?

carybakker: scratch that - the impl is here, and Data32 calls it, instead of the impl being in Data32...

carybakker (approved):
LGTM

the-real-adammork:
@carybakker  Thank you for your feedback and changes, it made this PR much better.

PR #113, opened 2022-02-15
Serialize execution of integration tests for supported protocols

the-real-adammork:
Soundtrack of this PR: [FIXATE - RIPGROOVE ReFIX](https://soundcloud.com/fixate_uk/ripgroove-refix)

### Motivation

Some integration tests fail from race conditions caused by running them back to back in short succession. 

### In this PR
* Add convenience function that iterates through supported protocols with a sleep interval between each test case.
* Convert `MobileCoinClientIntTests` and `MobileCoinClientPublicApiIntTests` to use new convenience function.

### DIFF
--- a/MobileCoin.podspec
+++ b/MobileCoin.podspec
@@ -58,7 +58,7 @@ Pod::Spec.new do |s|
       "Sources/Network/*.{h,m,swift}",
     ]
 
-    subspec.dependency "LibMobileCoin/Core", "~> 1.2.0-pre7"
+    subspec.dependency "LibMobileCoin/Core", "1.2.0-pre7"
 
     subspec.dependency "gRPC-Swift"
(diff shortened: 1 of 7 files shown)


MobileCoin.podspec
```diff
@@ -58,7 +58,7 @@ Pod::Spec.new do |s|
       "Sources/Network/*.{h,m,swift}",
     ]
 
-    subspec.dependency "LibMobileCoin/Core", "~> 1.2.0-pre7"
```
the-real-adammork, line 61: @carybakker 

This is why your CI tests were failing. the `~> `1.2.0-pre7` means choose a version equal to or greater than specified, but within the same major version (excludes >= 2.x.x).

I changed to comma syntax which is same as "version equals".

carybakker (approved):
LGTM - verified integration tests passing

PR #114, opened 2022-02-16
Internal Load Balancer

carybakker:

### Motivation

load balancer task
### In this PR
* adds url load balancing support

[iOS Implement Internal Load Balancer](https://app.asana.com/0/1201602719155430/1200761120709350)

### Future Work
* Additional tests for load balancer logic
* Adding protocol abstraction for load balancer impl


Sources/Network/HTTPS/HttpProtocolConnectionFactory.swift
```diff
@@ -19,7 +19,7 @@ class HttpProtocolConnectionFactory: ProtocolConnectionFactory {
     ) -> ConsensusHttpConnection {
         ConsensusHttpConnection(
                         config: config,
-                        requester: RestApiRequester(requester: requester, baseUrl: config.url),
+                        requester: RestApiRequester(requester: requester, baseUrl: config.urlLoadBalancer.nextUrl()),
```
the-real-adammork: i think this is a problem because you rotate to "nextUrl" to set baseUrl, but then the config gets passed down further the chain, and will have nextUrl called again in the Inner.init(...), possibly leading to a mismatch

Sources/Network/GRPC/GrpcChannelManager.swift
```diff
@@ -13,7 +13,7 @@ final class GrpcChannelManager {
 
     func channel(for config: ConnectionConfigProtocol) -> GRPCChannel {
         let wrappedTrustRoots = config.trustRoots[.grpc] as? WrappedNIOSSLCertificate
-        return channel(for: config.url, trustRoots: wrappedTrustRoots?.trustRoots)
+        return channel(for: config.currentUrl, trustRoots: wrappedTrustRoots?.trustRoots)
```
the-real-adammork: thinking this could be another place where a mismatch could occur

Sources/Network/NetworkConfig.swift
```diff
@@ -5,20 +5,23 @@
...
     }
 
-    let consensusUrl: ConsensusUrl
-    let fogUrl: FogUrl
-
     private let attestation: AttestationConfig
+    private let consensusUrlLoadBalancer: RandomUrlLoadBalancer<ConsensusUrl>
+    private let fogUrlLoadBalancer: RandomUrlLoadBalancer<FogUrl>
+    
+    var fogUrlsDescription: String {
```
the-real-adammork: don't think these should be member variables on this struct if just for debugging.

the-real-adammork:
Im having concern that were managing the rotation to far into the code-path for "creating" a new connection, and this could cause mismatch because we have to manage multiple downstream usage points.

One of my original ideas was to put the rotation in `HttpProtocolConnectionFactory`/`GrpcProtocolConnectionFactory` above the init for each particular connection. We know 100% this code path only gets called once each time a new connection is created. 

Also, we could make this change lower impact by passing the RandomLoadBalancer in as a separate parameter from the config object. So the config remains mostly the same as it is now and we don't have to propagate all the changes in at each of its usage sites.

Ill work on a prototype implementation for my idea

the-real-adammork:
suggested changes here - https://github.com/mobilecoinofficial/MobileCoin-Swift/pull/115

Sources/Network/NetworkConfig.swift
the-real-adammork: can we move these string constructions to in-line variables at the call-site ?

Sources/Common/Errors.swift
```diff
@@ -6,6 +6,20 @@
 
 import Foundation
 
+public struct UnexpectedError: Error {
```
the-real-adammork: Id prefer not to add a new generic error type, also not seeing it used anywhere so maybe its already been replaced ?

Sources/Network/Connection/Connections/BlockchainConnection.swift
```diff
@@ -11,37 +11,38 @@ final class BlockchainConnection:
...
     ) {
         self.httpFactory = httpFactory
         self.grpcFactory = grpcFactory
         self.config = config
         self.targetQueue = targetQueue
-
+        
         super.init(
             connectionOptionWrapperFactory: { transportProtocolOption in
+                let rotatedConfig = config.blockchainConfig()
```
the-real-adammork, line 30: agree that function syntax is more readable and correct for whats happening

the-real-adammork (commented):
Two comments left in-line, then I think we should do some real-world testing to make sure the rotation works. Im ok having this submitted as screenshots from proxyman or similar. Then I think its good to go.

Sources/Network/NetworkConfig.swift
carybakker: sure - good idea, will update

carybakker: the URLs are local to the load balancers, which are private members of the NetworkConfig - so the URLs are not accessible...  Will ping you in Slack to discuss how to expose the info for logging, if we want to.

carybakker: Updated NetworkConfig to expose fog and consensus urls (to not expose the load balancers themselves), where the MobileCoinClient extension builds the descriptive string for logging...

Sources/MobileCoinClient.swift
```diff
@@ -237,18 +237,26 @@ public final class MobileCoinClient {
...
+        }
+        
+        var consensusUrlsDescription: String {
+            "\(config.networkConfig.consensusUrls)"
+        }
+
         return """
-            Consensus url: \(config.networkConfig.consensusUrl.url)
-            Fog url: \(config.networkConfig.fogUrl.url)
+            Consensus urls: \(consensusUrlsDescription)
```
the-real-adammork:
```suggestion
            Consensus urls: \(config.networkConfig.consensusUrls)
```

Sources/MobileCoinClient.swift
```diff
@@ -237,18 +237,26 @@ public final class MobileCoinClient {
...
+            Fog urls: \(fogUrlsDescription)
```
the-real-adammork:
```suggestion
            Fog urls: \(config.networkConfig.fogUrls)
```

Sources/Common/Errors.swift
the-real-adammork: remove this error ?

Sources/MobileCoinClient.swift
```diff
@@ -237,18 +237,26 @@ public final class MobileCoinClient {
 extension MobileCoinClient {
     private static func configDescription(accountKey: AccountKeyWithFog, config: Config) -> String {
         let fogInfo = accountKey.fogInfo
+        var fogUrlsDescription: String {
```
the-real-adammork: thought we weren't doing this here ?

Sources/Network/Connection/Connections/ConsensusConnection.swift
```diff
@@ -32,35 +32,54 @@ final class ConsensusConnection:
...
                             rng: rng,
                             rngContext: rngContext))
                 }
             },
-            transportProtocolOption: config.transportProtocolOption,
+            transportProtocolOption: config.consensusConfig().transportProtocolOption,
             targetQueue: targetQueue)
     }
+    
+    func completionResultIntercept(
```
the-real-adammork: lets factor this out into one place.

you could make this a function on Connection.swift that returns a closure with a signature same as this function, then at the call-site you dont need the closure syntax

Tests/Integration/IntegrationTestFixtures.swift
```diff
@@ -84,6 +84,41 @@ extension IntegrationTestFixtures {
...
+            fogReportAttestation: network.fogReportAttestation(),
+            transportProtocol: transportProtocol).get()
+    }
+    
+    static func createMobileCoinClientConfigWithPartialValidFogUrls(transportProtocol: TransportProtocol) throws -> MobileCoinClient.Config {
+        try MobileCoinClient.Config.make(
+            consensusUrls: [network.consensusUrl],
+            consensusAttestation: network.consensusAttestation(),
+            fogUrls: [network.fogUrl, "fog://invalid.mobilecoin.com"],
+//            fogUrls: ["fog://invalid.mobilecoin.com"],
```
the-real-adammork: move this string into one of the fixtures files as a constant

Sources/Network/Connection/Connection.swift
```diff
@@ -34,6 +42,22 @@ class Connection<GrpcService: ConnectionProtocol, HttpService: ConnectionProtoco
...
+    
+    func rotateURLOnError<T>(
+        _ completion: @escaping (Result<T, ConnectionError>) -> Void
+    ) -> (Result<T, ConnectionError>) -> Void {
+        { [weak self] result in
+            switch result {
+            case .success:
+                completion(result)
+            case .failure:
+                logger.debug("ConsensusConnection - rotating config on error")
```
the-real-adammork: fyi, you don't need to put the filename in the logging statement, its included as a default parameter this is from the implementation:

```swift
    func debug(
        _ message: @autoclosure () -> String,
        metadata: @autoclosure () -> Logging.Logger.Metadata? = nil,
        source: @autoclosure () -> String? = nil,
        sensitive: Bool = false,
        logFunction: Bool = true,
        file: String = #file,
        function: String = #function,
        line: UInt = #line
    ) {
        debug(
            Message(stringLiteral: message()),
            metadata: metadata(),
            source: source(),
            sensitive: sensitive,
            logFunction: logFunction,
            file: file,
            function: function,
            line: line)
    }
```

`#file` `#function` etc.. are special tokens that capture this info.

carybakker: ya, noticed that when reviewing the logs and meant to remove - tx for reminder.

Tests/Integration/IntegrationTestFixtures.swift
carybakker: yup - the tests were WIP, will do... tx

carybakker:
Closing this PR to replace with new PR after squashing to rebase.

PR #115, opened 2022-02-16
Move rotation of URL upstream in the init path

the-real-adammork:
Has the benefit of preserving immutable data structures in more code, and isolates the rotation to the point at which a new connection is created for a given service. Seems like a lower impact change


the-real-adammork:
comparison with master - https://github.com/mobilecoinofficial/MobileCoin-Swift/compare/master...the-real-adammork:task/ios-load-balancer-squashed-low-impact?expand=1

PR #120, opened 2022-03-11
Version lock gRPC and SwiftNIO

the-real-adammork:
Soundtrack of this PR: [Cherrelle - Saturday](https://youtu.be/efLcgUQmyT8)

### Motivation

Version lock GRPC and SwiftNIO dependencies to preserve network robustness when connections are idle. Integrating apps may see this error after a connection goes idle `NIOCore/EventLoopFuture.swift:429: Precondition failed: leaking an unfulfilled Promise`.

### In this PR
* Version locking for GRPC and SwiftNIO dependencies, latest versions cause errors
* Revert changes that were implemented based on higher-version's deprecations

### Future Work
* Fix idle connection memory leak for newer versions of GRPC & SwiftNIO


carybakker (approved):
LGTM

PR #127, opened 2022-03-28
Recoverable Transaction History - Part 1 & 2

the-real-adammork:
Soundtrack of this PR: [Roy Ayers - Oh What a Lonely Feeling](https://youtu.be/-29dcLdhOp4)

### Motivation

Part 1 - Add support for Recoverable Transaction History. Part 2 will be branched off this PR, and then merged in before final merge. 

Part 2 - Add support for Recoverable Transaction History. Higher level RTH implementations.

### In this PR
* Memo & Data Primitives 
* AddressHash
* FFI Wrappers
* BlockVersion enum
* Updated libmobilecoin-ios-artifacts generated
* Updates to TxOut definitions
* Updates to TransactionBuilder
* TxOutMemoParser
* Updated libmobilecoin-ios-artifacts generated

### Future Work
* Dynamically pull `block_version` from the ledger and store for use in the TxBuilder

`mobilecoin` https://github.com/mobilecoinfoundation/mobilecoin/pull/1701

`libmobilecoin-ios-artifacts` https://github.com/mobilecoinofficial/libmobilecoin-ios-artifacts/pull/20

MCIP 004 - https://github.com/mobilecoinfoundation/mcips/blob/main/text/0004-recoverable-transaction-history.md
MCIP 026 - https://github.com/mobilecoinfoundation/mcips/blob/main/text/0024-increase-memo-field-size.md


Sources/LibMobileCoin/Data16+withMcMutableBuffer.swift
```diff
@@ -0,0 +1,68 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+import LibMobileCoin
+
+extension Data16 {
```
dolanbernard, line 8: Not an error, but it would be nice if we could get rid of some of the repeated code in this type and Data64 and Data66. Maybe you could maintain type safety by creating a type for a buffer of a specified size and then have these three extend it but instantiate with a fixed size.

Sources/Utils/Data/Data16.swift
```diff
@@ -0,0 +1,49 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+struct Data16 {
```
dolanbernard, line 7: Same as Data16+withMutableBuffer

dolanbernard (approved):
Looks good. One comment for consideration

Sources/Transaction/Memos/DestinationMemoDataUtils.swift
```diff
@@ -0,0 +1,186 @@
+//
```
carybakker: Filename is different than classname... has "DataUtils" suffix where the other memo util classes are just <memotype>Utils

Sources/Transaction/Memos/DestinationMemoDataUtils.swift
```diff
@@ -0,0 +1,186 @@
...
+            case .success(let bytes):
+                return bytes as Data16
+            case .failure(let error):
+                switch error.errorCode {
+                case .invalidInput:
+                    // Safety: This condition indicates a programming error and can only
+                    // happen if arguments to mc_memo_destination_memo_get_address_hash are
+                    // supplied incorrectly.
+                    logger.warning("error: \(redacting: error)")
+                    return Data16()
```
carybakker: Just wondering if we should make this return nil on error like the other util functions, as opposed to an empty Data16() for consistency?

Sources/Transaction/Memos/SenderMemoUtils.swift
```diff
@@ -0,0 +1,122 @@
...
+        memoData: Data64,
+        senderPublicAddress: PublicAddress,
+        receipientViewPrivateKey: RistrettoPrivate,
+        txOutPublicKey: RistrettoPublic
+    ) -> Bool {
+        memoData.asMcBuffer { memoDataPtr in
+            senderPublicAddress.withUnsafeCStructPointer { publicAddressPtr in
+                receipientViewPrivateKey.asMcBuffer { receipientViewPrivateKeyPtr in
+                    txOutPublicKey.asMcBuffer { txOutPublicKeyPtr in
+                        var matches = true
```
carybakker: We should probably default matches to false here...shouldn't change things if mc_memo_sender_memo_is_valid behaves properly and always sets it on .success, but just seems better to default to false when looking for a match in case of a bug in the called function.

Sources/Transaction/Memos/SenderWithPaymentRequestMemoUtils.swift
```diff
@@ -0,0 +1,158 @@
...
+        memoData: Data64,
+        senderPublicAddress: PublicAddress,
+        receipientViewPrivateKey: RistrettoPrivate,
+        txOutPublicKey: RistrettoPublic
+    ) -> Bool {
+        memoData.asMcBuffer { memoDataPtr in
+            senderPublicAddress.withUnsafeCStructPointer { publicAddressPtr in
+                receipientViewPrivateKey.asMcBuffer { receipientViewPrivateKeyPtr in
+                    txOutPublicKey.asMcBuffer { txOutPublicKeyPtr in
+                        var matches = true
```
carybakker: We should probably default matches to false here...shouldn't change things if mc_memo_sender_with_payment_request_memo_is_valid behaves properly and always sets it on .success, but just seems better to default to false when looking for a match in case of a bug in the called function.

Sources/Account/AddressHash.swift
```diff
@@ -0,0 +1,41 @@
...
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+import LibMobileCoin
+
+struct AddressHash {
+    let data16: Data16
+    
+    var hexBytes: String {
```
carybakker: Minor nit, but the name 'hexBytes' makes it seem like this value is raw data bytes in some form, versus a string... can we rename to hexString?

carybakker (approved):
Approving with comments / suggestions

Sources/LibMobileCoin/Data16+withMcMutableBuffer.swift
the-real-adammork: Looked into doing this, and using a class I don't like because we loose the properties of a struct. Theres sort of way todo it with structs, but it would require I rewrite a bunch of other protocols and downgrade variable access levels which would end up being a high-impact change because these protocols are used widely in other places of the project. I think for now the duplication is acceptable. I can revisit in the future if I think of a third, better, option.

Sources/Transaction/Memos/SenderMemoUtils.swift
the-real-adammork: good point, updated

Sources/Transaction/Memos/SenderWithPaymentRequestMemoUtils.swift
the-real-adammork: good point, updated

Sources/Account/AddressHash.swift
the-real-adammork: good point, updated

Sources/Transaction/Memos/DestinationMemoDataUtils.swift
the-real-adammork: the other "address hash" ffi functions always return a value because its just some computation on bytes, so in that spirit im making this always return a valid value. It makes sense because it will always succeed, and making it non-nillable simplifies a bunch of other code.

Sources/Transaction/Memos/DestinationMemoDataUtils.swift
the-real-adammork: updated

PR #128, opened 2022-03-28
Recoverable Transaction History - Part 2: Higher Level Implementations (TxOut, TransactionBuilder, TxOutMemoParser)

the-real-adammork:
Soundtrack of this PR: [Roy Ayers - Oh What a Lonely Feeling](https://youtu.be/-29dcLdhOp4)

### Motivation

Part 2 - Add support for Recoverable Transaction History. Higher level RTH implementations.

### In this PR
* Updates to TxOut definitions
* Updates to TransactionBuilder
* TxOutMemoParser
* Updated libmobilecoin-ios-artifacts generated

### TODO Before Merge
* make static libraries compatible with latest `mobilecoin` master

### Future Work
* Dynamically pull `block_version` from the ledger and store for use in the TxBuilder

### Related PRs

`mobilecoin` https://github.com/mobilecoinfoundation/mobilecoin/pull/1701

`libmobilecoin-ios-artifacts` https://github.com/mobilecoinofficial/libmobilecoin-ios-artifacts/pull/20

MCIP 004 - https://github.com/mobilecoinfoundation/mcips/blob/main/text/0004-recoverable-transaction-history.md
MCIP 026 - https://github.com/mobilecoinfoundation/mcips/blob/main/text/0024-increase-memo-field-size.md


Sources/Account/Account+TransactionOperations.swift
```diff
@@ -162,6 +169,7 @@ extension Account {
 
         func prepareDefragmentationStepTransactions(
             toSendAmount amountToSend: UInt64,
+            recoverableMemo: Bool,
```
carybakker, line 172: wdyt of renaming `recoverableMemo` to something like `isRecoverable` ?    While the memo is meant to provide info to recover transactions, the memo itself isn't really what's being recovered... Since we're talking about recoverable transactions, just having a flag that refers to recoverability should be enough, right?   Also, it'll match the naming convention for boolean values...  Could be one of:
- isRecoverable
- useRecoverability  (maybe this is better, b/c name refers to transactions (plural), but isRecoverable is singular)

Main point is that recoverableMemo sounds like a memo as opposed to a boolean flag.

Sources/Transaction/Memos/SenderMemo.swift
```diff
@@ -0,0 +1,44 @@
...
+
+struct SenderMemo {
+    let memoData: Data64
+    let addressHash: AddressHash
+}
+
+struct RecoverableSenderMemo {
+    let memoData: Data64
+    let addressHash: AddressHash
+    let accountKey: AccountKey
```
dolanbernard: If TxOuts and memos are going to be cached, then this account key will too. We need to make sure AccountKeys won't be stored insecurely.

Sources/Transaction/Memos/DestinationMemo.swift
```diff
@@ -0,0 +1,51 @@
...
+    let addressHash: AddressHash
+    let numberOfRecipients: PositiveUInt8
+    let fee: UInt64
+    let totalOutlay: UInt64
+}
+
+struct RecoverableDestinationMemo {
+    let memoData: Data64
+    let addressHash: AddressHash
+    let accountKey: AccountKey
```
dolanbernard: Same as SenderMemo

Sources/Transaction/Memos/SenderWithPaymentRequestMemo.swift
```diff
@@ -0,0 +1,52 @@
...
+struct SenderWithPaymentRequestMemo {
+    let memoData: Data64
+    let addressHash: AddressHash
+    let paymentRequestId: UInt64
+}
+
+struct RecoverableSenderWithPaymentRequestMemo {
+    let memoData: Data64
+    let addressHash: AddressHash
+    let accountKey: AccountKey
```
dolanbernard: Same as SenderMemo

dolanbernard (approved):
Looks good as long as AccountKey comments aren't an issue

Sources/Transaction/Memos/RecoverableMemo.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    init(decryptedMemo data: Data66, accountKey: AccountKey, txOutKeys: TxOut.Keys) {
+        guard let memoData = Data64(data[2...]) else {
+            logger.warning("Memo data type unavailable")
+            self = .notset
+            return
+        }
+        let typeBytes = data[..<2]
+        
+        switch typeBytes.hexEncodedString() {
```
carybakker, line 27: We should just be mapping the typeBytes to an integer value, and comparing to values defined numerically with constants (UInt16)... no need to convert bytes to string

Sources/Transaction/Memos/RecoverableMemo.swift
```diff
@@ -0,0 +1,74 @@
...
+        case Types.UNUSED:
+            self = .unused
+        default:
+            logger.warning("Memo data type unknown")
+            self = .notset
+        }
+    }
+    
+    struct Types {
+        static let SENDER = "0100"
```
carybakker, line 46: e.g. 0x0100, 0x0101, 0x200, etc..

Sources/Transaction/Memos/SenderMemo.swift
```diff
@@ -0,0 +1,44 @@
...
+        }
+        return SenderMemo(memoData: memoData, addressHash: addressHash)
+    }
+}
+
+extension RecoverableSenderMemo: Hashable { }
+
+extension RecoverableSenderMemo: Equatable {
+    static func == (lhs: Self, rhs: Self) -> Bool {
+        return lhs.memoData == rhs.memoData
```
carybakker, line 42: Is it really valid to only compare memoData?   Practically speaking, I think it's fine, but it just seems strange that other attributes of the recoverable can be different but they can still be considered equal.

Sources/Transaction/Memos/RecoverableMemo.swift
```diff
@@ -0,0 +1,74 @@
...
+
+enum RecoverableMemo {
+    case notset
+    case unused
+    case sender(RecoverableSenderMemo)
+    case destination(RecoverableDestinationMemo)
+    case senderWithPaymentRequest(RecoverableSenderWithPaymentRequestMemo)
+    
+    init(decryptedMemo data: Data66, accountKey: AccountKey, txOutKeys: TxOut.Keys) {
+        guard let memoData = Data64(data[2...]) else {
```
carybakker, line 22: Given that the source data is Data66, we're guaranteed to get a Data64 out of [2...], but not sure if we should change anything here since the subrange returns an optional and we don't like forced unwrap....  But we definitely don't need the logger warning and the assignment to .notset since this code is unreachable.. maybe just a comment?

Sources/Transaction/Memos/SenderMemo.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+
+struct SenderMemo {
+    let memoData: Data64
```
carybakker, line 9: One thing I've been wondering - do we really need to copy and keep around the raw memo data once the SenderMemo has been extracted?    Seems like there's maybe no need for it in the direct memo classes.

Also, it's not really necessary to copy the data into the recoverable to parse out the values... not sure how important this may or may not be (will depend on the number of transactions we're talking about having to scan for a user), so may be over-optimizing...

Sources/Account/Account+TransactionOperations.swift
```diff
@@ -81,10 +83,12 @@ extension Account {
                 transactionPreparer.prepareTransaction(
                     inputs: txOutsToSpend,
                     recipient: recipient,
+                    memoType: memoType,
                     amount: amount,
                     fee: fee,
                     tombstoneBlockIndex: tombstoneBlockIndex,
-                    completion: completion)
+                    blockVersion: MobileCoinClient.latestBlockVersion,
+                    completion: completion) // - TODO
```
carybakker: What's this TODO for?

Sources/Ledger/TxOutUtils.swift
```diff
@@ -99,6 +99,51 @@ enum TxOutUtils {
         }
     }
 
+    static func decryptEMemoPayload(
```
carybakker, line 102: Not sure I'm crazy about the "EMemo" name here.. guessing the E is meant to stand for Encrypted, but it's not really necessary since the function name starts with decrypt, and the parameter name itself is "encryptedMemo"... 

Just trying to remove things that make people (like me lol) wonder "What's an EMemo?" when it's not really a type or a thing...

carybakker: Ahhh, I see the name ..._e_memo_... in the wrapped function call...  But still, if it's wrapping that call abstraction, no real need to percolate the name up to the swift function.

Sources/Ledger/TxOut.swift
```diff
@@ -76,8 +84,31 @@ extension TxOut {
     }
 }
 
+extension TxOut {
+//    // TODO use result type or optional ?
```
carybakker: Does this need to be removed or uncommented?

Sources/Transaction/Memos/SenderWithPaymentRequestMemo.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+// TODO - Decide how to manage isValid for memo's where senderPublicAddress not nearby ?
```
carybakker: reminder to revisit all the TODOs

Sources/Transaction/Memos/RecoverableMemo.swift
```diff
@@ -0,0 +1,74 @@
...
+        switch typeBytes.hexEncodedString() {
+        case Types.SENDER_WITH_PAYMENT_REQUEST:
+            let memo = RecoverableSenderWithPaymentRequestMemo(memoData, accountKey: accountKey, txOutPublicKey: txOutKeys.publicKey)
+            self = .senderWithPaymentRequest(memo)
+        case Types.SENDER:
+            let memo = RecoverableSenderMemo(memoData, accountKey: accountKey, txOutPublicKey: txOutKeys.publicKey)
+            self = .sender(memo)
+        case Types.DESTINATION:
+            let memo = RecoverableDestinationMemo(memoData, accountKey: accountKey, txOutKeys: txOutKeys)
+            self = .destination(memo)
```
carybakker, line 36: One thought here... 

Each of the "Revoverable<Memotype>()" functions can possibly return nil if the data is not valid...  

When that happens, we'll still be returning a non-nil RecoverableDemo enum for one of the valid types, but the memo itself will be nil...   

Almost wondering if we should add an 'invalid' case to use when the memo isn't parseable b/c of bad data?

carybakker: As a logical comparison, if the decryption fails, we set the value to .notSet in TxOutMemoParser...

carybakker (approved):
Approved with comments / feedback
Need to look over test scenarios / coverage a bit more still, focused mostly on code-code.

Sources/Transaction/Memos/SenderMemo.swift
the-real-adammork: added private access

Sources/Transaction/Memos/DestinationMemo.swift
the-real-adammork: added private access

Sources/Transaction/Memos/SenderWithPaymentRequestMemo.swift
the-real-adammork: added private access

Sources/Account/Account+TransactionOperations.swift
the-real-adammork: isRecoverable makes sense if its a property of an object, I think "recoverableMemo: Bool" gets the point across without the need for unecc. extra words

Sources/Transaction/Memos/SenderMemo.swift
the-real-adammork: Agree this would be better but could not find a way todo this that would be simpler than string conversion

Sources/Transaction/Memos/RecoverableMemo.swift
the-real-adammork: Agree this would be better but could not find a way todo this that would be simpler than string conversion

Sources/Transaction/Memos/RecoverableMemo.swift
the-real-adammork: see above, unable todo this better than using strings.

Sources/Transaction/Memos/SenderMemo.swift
the-real-adammork: everything else is derivative or extra stuff used to recover, I think for now it makes sense if it needs to change in the future for some application I will revisit.

Sources/Transaction/Memos/RecoverableMemo.swift
the-real-adammork: updated, made it logger.fatalError()

Sources/Transaction/Memos/SenderMemo.swift
the-real-adammork: first question -  ya Im not sure we need the memoData, I guess this is open-ended for me until we add the higher level abstractions

second part - I like wrapping it the enum, it adds the extra piece of state that the memo is yet to be recovered, Im thinking of a high-level api that emits all of these as "recoverable" then depending on the access needs we iterate through and "recover" as needed.

Sources/Account/Account+TransactionOperations.swift
the-real-adammork: removed, was addressed

Sources/Ledger/TxOutUtils.swift
the-real-adammork: i like EMemo its camel case for e_memo

Sources/Transaction/Memos/SenderWithPaymentRequestMemo.swift
the-real-adammork: removed its been addressed

Sources/Ledger/TxOut.swift
the-real-adammork: removed its been addressed. Added eMemo to the translation constructor

Sources/Transaction/Memos/RecoverableMemo.swift
the-real-adammork: the constructors always return valid data. At one point they were nillable but I changed that by assuming all "addressHash" ffi functions return data, which is true, they generally always succeed with some sort of value because its just a computation on bytes.

PR #130, opened 2022-04-29
Update Podfile for Xcode13 support

the-real-adammork:
Soundtrack of this PR: [Aril Brikha - Aqua](https://www.youtube.com/watch?v=ZgWFhwhhWEA)

### Motivation

The currently linked version of SwiftNIO is incompatible with Xcode 13. This has a been annoyance for many working with the SDK as it forced some to downgrade.

### In this PR
* Dependency updates to support Xcode 13

PR #131, opened 2022-04-29
Prevent client showing wrong balance when Fog and/or Consensus is out of Sync

the-real-adammork:
Soundtrack of this PR: [Things From The Basement - No One Belongs Here More Than You (Map.ache Remix)](https://www.youtube.com/watch?v=-MCd7i2WvWE)

### Motivation

Fog View, Ledger, and Consensus can get out of sync during blockchain deployments. This would cause the SDK to temporarily show an incorrect balance. There are some heuristics we can use to detect this situation. We can then throw an error instead of returning a balance number.

### In this PR
* Add FogSyncChecker class which has ways to read & write the "highest known block" for View, Ledger and Consensus
* Wrap it a serial write & concurrent read dispatch lock to ensure thread safety
* Unit tests
* Integration Tests
* Includes the Xcode 13 changes, Ill merge the changes for that in this isolated PR first https://github.com/mobilecoinofficial/MobileCoin-Swift/pull/130


Sources/Fog/FogSyncChecker.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+    
+    var currentBlockIndex: UInt64 {
+        min(ledgersHighestKnownBlock, viewsHighestKnownBlock)
+    }
+    
+    private var viewLedgerOutOfSync: Bool {
+        abs(
+            Int64(
+                max(viewsHighestKnownBlock, ledgersHighestKnownBlock)
```
carybakker: using possessive 's' is a bit strange in variable names b/c no apostrophe which makes it seem like it's meant to be a plural, would prefer highestKnownViewBlock highestKnownLedgerBlock

Sources/Fog/FogSyncChecker.swift
```diff
@@ -0,0 +1,86 @@
...
+            Int64(consensusHighestKnownBlock &- currentBlockIndex) >= fogSyncThreshold
+    }
+}
+
+class FogSyncChecker: FogSyncCheckable {
+    var viewsHighestKnownBlock: UInt64 = 0
+    var ledgersHighestKnownBlock: UInt64 = 0
+    var consensusHighestKnownBlock: UInt64 = 0
+    
+    let fogSyncThreshold: UInt64 = 10
```
carybakker: Is it really true that we're in sync when < X blocks behind?  Not sure I intuitively understand this / how the threshold makes sense (as a layperson, I'm thinking it's either in sync or not, meaning I'd expect us to always want to use a threshold of 0), but maybe I'm missing some context on how this is supposed to work

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -0,0 +1,266 @@
...
+}
+
+class MockFogSyncChecker: FogSyncCheckable {
+    var viewsHighestKnownBlock: UInt64
+    var ledgersHighestKnownBlock: UInt64
+    var consensusHighestKnownBlock: UInt64
+
+    let fogSyncThreshold: UInt64
+    
+    static let threshold: UInt64 = 10
```
carybakker: Since this is a UInt, the user *could* set a threshold to 0... meaning that things will always show as out of sync if the block counts match, but if we'll ALWAYS need to use a non-zero threshold for some reason, then maybe we disallow it somehow by throwing an exception if the value is ever set to zero?

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+    
+    func viewLedgerAndConsensusAtThreshold(
+        transportProtocol: TransportProtocol,
+        expectation expect: XCTestExpectation
+    ) throws {
+        let viewIndex: UInt64 = 99
+        let ledgerIndex = viewIndex + MockFogSyncChecker.threshold + 1
+        let consensusIndex = viewIndex + MockFogSyncChecker.threshold / 2
```
carybakker, line 166: Isn't this really within threshold,  instead of at threshold?  At threshold makes me think they should be both set to viewIndex + threshold

Sources/Fog/FogSyncChecker.swift
```diff
@@ -0,0 +1,86 @@
...
+        abs(
+            Int64(
+                max(viewsHighestKnownBlock, ledgersHighestKnownBlock)
+                &- min(viewsHighestKnownBlock, ledgersHighestKnownBlock)
+            )
+        ) >= fogSyncThreshold
+    }
+    
+    private var consensusOutOfSync: Bool {
+        consensusHighestKnownBlock > currentBlockIndex &&
```
carybakker, line 58: Do you really need the first part of this check?  It should be sufficient to subtract and check the threshold, right?

Sources/Fog/FogSyncChecker.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+        return .success(())
+    }
+    
+    var currentBlockIndex: UInt64 {
+        min(ledgersHighestKnownBlock, viewsHighestKnownBlock)
+    }
+    
+    private var viewLedgerOutOfSync: Bool {
+        abs(
```
carybakker: Since you're doing a max and min within this, do you really need the abs since you're guaranteed that the subtracted value will always be <= value it's subtracted from?

Sources/Fog/FogSyncChecker.swift
```diff
@@ -0,0 +1,86 @@
...
+//
+
+import Foundation
+
+protocol FogSyncCheckable {
+    var viewsHighestKnownBlock: UInt64 { get }
+    var ledgersHighestKnownBlock: UInt64 { get }
+    var consensusHighestKnownBlock: UInt64 { get }
+    var currentBlockIndex: UInt64 { get }
+    var fogSyncThreshold: UInt64 { get }
```
carybakker: Threshold doesn't always clearly indicate whether equality is acceptable.  I still catch myself using it in names too, but thinking something more semantic would be better, like 'maxAllowedBlockDelta', which clearly indicated equality via 'allowed', and delta indicating the difference between blocks (or some similar descriptive name attempt).

Also, it seems like we're able to set a threshold of 0, but a threshold of 0 will always return out of sync because we're checking for out of sync using <=.

It's kind of 6 versus a half-dozen, where we could set the threshold to 1 in order to require that the block values match exactly (but I'm not sure if we need to support that).

Just thinking that maybe the name of this variable could indicate what to do when the delta is equal to the threshold, and also wondering why we need to support saying we're in sync when we're just close (that's just my lack of understanding of the business logic).. maybe it's because the value changes so rapidly that we need some kind of allowed delta or we'll never be fully in sync?

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -0,0 +1,266 @@
...
+    
+    func testFogBehindConsensusWithinThreshold() throws {
+        try testSupportedProtocols(description: description) {
+            try fogBehindConsensusWithinThresholdAheadConsensus(
+                    transportProtocol: $0,
+                    expectation: $1)
+        }
+    }
+    
+    func fogBehindConsensusWithinThresholdAheadConsensus(
```
carybakker: is there supposed to be the "AheadConsensus" suffix to this test function name?

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -0,0 +1,266 @@
...
+            try fogBehindConsensusAtThreshold(transportProtocol: $0, expectation: $1)
+        }
+    }
+    
+    func fogBehindConsensusAtThreshold(
+        transportProtocol: TransportProtocol,
+        expectation expect: XCTestExpectation
+    ) throws {
+        let fogIndex: UInt64 = 60
+        let consensusIndex = fogIndex + MockFogSyncChecker.threshold - 1
```
carybakker: Doesn't 'AtThreshold' mean that this line should be `let consensusIndex = fogIndex + MockFocSyncChecker.threshold` ?

carybakker (commented):
Just a few quick things in the comments...

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -0,0 +1,266 @@
...
+            try fogBehindConsensusOverThreshold(transportProtocol: $0, expectation: $1)
+        }
+    }
+    
+    func fogBehindConsensusOverThreshold(
+        transportProtocol: TransportProtocol,
+        expectation expect: XCTestExpectation
+    ) throws {
+        let fogIndex: UInt64 = 410
+        let consensusIndex = fogIndex + MockFogSyncChecker.threshold
```
carybakker: I would have expected this to mean "AtThreshold" instead of "OverThreshold" where you're adding the threshold value,  so when the sync checker is subtracting and getting the delta, it will be == threshold (i.e. at threshold), no?

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -0,0 +1,266 @@
...
+        )
+    }
+    
+    func testViewLedgerAndConsensusAtThreshold() throws {
+        try testSupportedProtocols(description: description) {
+            try viewLedgerAndConsensusAtThreshold(transportProtocol: $0, expectation: $1)
+        }
+    }
+    
+    func viewLedgerAndConsensusAtThreshold(
```
carybakker: Am I wrong that the test method names don't seem to match the data?   For example, for this test method name, I would have expected `ledgerIndex == consensusIndex == viewIndex + MockFogSyncChecker.threshold`

Sources/Fog/FogSyncChecker.swift
the-real-adammork: I like how the first word disambiguates tho

Sources/Fog/FogSyncChecker.swift
the-real-adammork: threshold is in the requirements for this feature

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: this is a unit test, but the other place we define this var its marked with access level internal so SDK consumers won't be able to see it.

Sources/Fog/FogSyncChecker.swift
the-real-adammork: We get number overflow issues when subtracting UInt64 - UInt64 when the second is larger, causes integer overflow exception. Not sure of the exact mechanics of this operation, but ensuring the subtraction result is positive avoids this issue, and the check is a valid short-circuit. 

thinking about it, I should annotate this line with a comment, like above, to disambiguate this for future readers.

Sources/Fog/FogSyncChecker.swift
the-real-adammork: same as above, avoids the possibility of a negative result in one of the subtraction operations, I will add a comment to disambiguate this for future readers.

Sources/Fog/FogSyncChecker.swift
the-real-adammork: Good point about the naming, lets think of a better way to be more descriptive of the delta/threshold allowed. 

As for setting to 0, I don't think thats a major concern because this data structure isnt exposed outside of the SDK module, and not really intended for general consumption (tampering) by them, but maybe Im missing something with the access controls that could allow such a case.

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: I might need todo some renaming one way or the other here because we're able to get the "consensus **block-count**" (1-N) from the network, but we store it as an **index** (0 - N-1) in the FogSyncChecker.  So in the unit test we -1 to make the block count an index, and all the math assumes we are dealing with indexes, but could just as easily store the consensusBlockCount in the fogSyncChecker and modify the `inSync` calculations with `-1`

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: good point let me double check

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: ya this needs to be adjusted

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork:
```suggestion
    func fogBehindConsensusButWithinThreshold(
```

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
carybakker: I was thinking "user" in terms of the user of the code - even if it's us...   It's possible that logically it would never make sense to set threshold to zero - I'll review the requirements for the feature.

Sources/Fog/FogSyncChecker.swift
carybakker: Got it...

Sources/Fog/FogSyncChecker.swift
carybakker: sorry - what do you mean first work?

the-real-adammork: corrected:

> I like how the first word disambiguates tho

```
viewXXX
ledgerXXX
consensusXXX
```

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: I think this is a solution looking for a problem, but we could use `PositiveUInt64`

Sources/Fog/FogSyncChecker.swift
carybakker: maybe just remove the "s" to make them "viewHighestBlock" and "ledgerHighestKnownBlock"?  The only thing bad about that is that "view" is both a noun and a verb, but if it's consistent with the rest of the code, that should be at least as ok as the code already is.

the-real-adammork:
realizing some of the code comments on the integration tests are wrong, updating now

the-real-adammork:
comments/naming updated

Sources/Fog/FogSyncChecker.swift
the-real-adammork: I like the readability of the possesive, lets revisit this type of situation when we work on the linting rules, thats one of the goals for this quarter and we might be able to enforce possesive/non-possesive w/ a grep rule in swiftlint.

Sources/Fog/FogSyncChecker.swift
the-real-adammork: updated to `maxAllowedDelta`, much better now

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: updated

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: updated

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -48,156 +51,162 @@ class FogSyncErrorConnectionIntTests: XCTestCase {
...
-            shouldSucceed: true
+            shouldSucceed: shouldSucceed
         )
     }
     
-    func testFogBehindConsensusWithinThreshold() throws {
+    func testFogBehindConsensusWithinDelta() throws {
         try testSupportedProtocols(description: description) {
-            try fogBehindConsensusWithinThresholdAheadConsensus(
+            try fogBehindConsensusWithinDeltaAheadConsensus(
```
carybakker:
```suggestion
            try fogBehindConsensusWithinDelta(
```

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -48,156 +51,162 @@ class FogSyncErrorConnectionIntTests: XCTestCase {
...
                     transportProtocol: $0,
                     expectation: $1)
         }
     }
     
-    func fogBehindConsensusWithinThresholdAheadConsensus(
+    func fogBehindConsensusWithinDeltaAheadConsensus(
```
carybakker:
```suggestion
    func fogBehindConsensusWithinDelta(
```

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
```diff
@@ -48,156 +51,162 @@ class FogSyncErrorConnectionIntTests: XCTestCase {
...
-    func testViewLedgerAndConsensusOverThreshold() throws {
+    func testViewLedgerAndConsensusOverDelta() throws {
         try testSupportedProtocols(description: description) {
-            try viewLedgerAndConsensusOverThreshold(transportProtocol: $0, expectation: $1)
+            try viewLedgerAndConsensusOverDelta(transportProtocol: $0, expectation: $1)
         }
     }
     
-    func viewLedgerAndConsensusOverThreshold(
+    func viewLedgerAndConsensusOverDelta(
```
carybakker, line 193: for all the viewLedgerAndConsensus* tests, the names read like "LedgerAndConsensus" are supposed to both be the same, is that not the case?

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: oh ya I see now this was a typo on my part

the-real-adammork: both should be fixed now

carybakker (approved):

Tests/Integration/Network/Connection/FogSyncErrorConnectionIntTests.swift
the-real-adammork: viewLedgerAndConsensus indicates they are all different values, fogAndConsensus indicates view == ledger

PR #132, opened 2022-05-03
Dynamically Fetch & Cache `BlockVersion` for Transaction Building

the-real-adammork:
Soundtrack of this PR: [Mac Dre - All Damn Day](https://www.youtube.com/watch?v=Jm1ToRW0ip8)

### Motivation

The `BlockVersion` is important for building transactions. It can be fetched at the same time as the fee, and is also returned in the `ProposeTxResponse`. We can use the existing "fee cache" to also cache the current `BlockVersion`.

### In this PR
* Change the name of `BlockchainFeeFetcher` to `BlockchainMetaFetcher` to encompass its new responsibilities
* Store `BlockVersion` in the cache the same way we do the fee as its also available in the `getLastBlockInfo` call
* Add a rule to invalidate the cache if the cached block version is different from the one in the `ProposeTxResponse`

### Future Work
*  In a future PR we should classify errors as retry-able and potentially restructure the public API to allow for the SDK to rebuild the transactions with the new BlockVersion. Right now the public api exposes a call `submitTransaction` but at that point the transaction is already built, and it may be difficult to deconstruct, rebuild, and re-submit. More investigation is needed to determine how best to enable automatic retry on `txFeeError` or "block version based errors".
* https://app.asana.com/0/1200318482173758/1202215095222003/f


voloshyn (approved):

Sources/Account/Account+TransactionOperations.swift
```diff
@@ -115,7 +132,7 @@ extension Account {
                 return
             }
 
-            feeFetcher.feeStrategy(for: feeLevel) {
+            metaFetcher.feeStrategy(for: feeLevel) {
```
carybakker, line 135: `metaFetcher.feeStragety` looks like it'll just use the cached value, if present... can it change or maybe not be fully up to date? 

Reason I'm asking is b/c metaFetcher.blockVersion` can trigger a `fetchMeta` which looks like it may also udpate feeStrategy...  

So seems like the fee from this call could be stale but used anyways.

If this is a problem, maybe we can flip the order of the calls (check block then get fee), so that the block number is checked first - which will update the fee in tandem, then the subsequent call to get the fee will be using the current value pulled by the fetchMeta call.

Sources/Ledger/BlockVersion.swift
```diff
@@ -4,27 +4,22 @@
...
-        version >= one
+        version >= versionOne
     }
     
     static var legacy: BlockVersion {
-        zero
+        Self.versionZero
     }
     
     static var minRTHEnabled: BlockVersion {
```
carybakker, line 22: I like the semantic naming of minRTHEnabled, but should we also add another var, like `currentBlockVersion` here and set it == `minRTHEnabled`, so that we don't have to make any changes in other files when we want to update the currently supported block version?    (i.e. the TxOut, Receipt, and Transaction+Fixtures files that each create their own `defaultBlockVersion` set to `minRTHEnabled` would each have to be updated, and if one is missed, it could be referencing the wrong block version)

Tests/Common/Fixtures/Ledger/TxOut+Fixtures.swift
```diff
@@ -37,7 +37,7 @@ extension TxOut.Fixtures {
 }
 
 extension TxOut.Fixtures.Default {
-    fileprivate static let blockVersion = BlockVersion.one
+    fileprivate static let defaultBlockVersion = BlockVersion.minRTHEnabled
```
carybakker, line 40: Move this to BlockVersion.swift?

Tests/Common/Fixtures/Transaction/Receipt+Fixtures.swift
```diff
@@ -36,7 +36,7 @@ extension Receipt.Fixtures {
 
 extension Receipt.Fixtures.Default {
 
-    fileprivate static let blockVersion = BlockVersion.one
+    fileprivate static let defaultBlockVersion = BlockVersion.minRTHEnabled
```
carybakker, line 39: Move this to BlockVersion.swift?

Tests/Common/Fixtures/Transaction/Transaction+Fixtures.swift
```diff
@@ -18,7 +18,7 @@ extension Transaction.Fixtures {
         let fee = Self.fee
         let tombstoneBlockIndex = Self.tombstoneBlockIndex
         let fogResolver: FogResolver
-        let blockVersion = Self.blockVersion
+        let blockVersion = Self.defaultBlockVersion
```
carybakker, line 21: Move this to BlockVersion.swift?

Tests/Common/Fixtures/Transaction/Transaction+Fixtures.swift
```diff
@@ -62,7 +62,7 @@ extension Transaction.Fixtures {
 }
 
 extension Transaction.Fixtures.BuildTx {
-    fileprivate static var blockVersion = BlockVersion.one
+    fileprivate static var defaultBlockVersion = BlockVersion.minRTHEnabled
```
carybakker, line 65: Move this to BlockVersion.swift?

Sources/Account/Account+TransactionOperations.swift
the-real-adammork: this change is rename so don't think it makes sense to refactor behavior if its already working and tested. Answering your question: feeStrategy() { ... }  has a closure so that it uses the cached value if its not-expired, and fetches the network if it is, then executes the closure.

Sources/Ledger/BlockVersion.swift
the-real-adammork: Not sure we will want to adjust the unit tests as youve suggested when we add more BlockVersion support. Some tests only make sense on certain block versions, other probably need to be tested on all block versions. I will add an asana ticket for this task and bundle it into the work I do when we start supporting/testing BlockVersion == 2

Tests/Common/Fixtures/Ledger/TxOut+Fixtures.swift
the-real-adammork: for right now this works, in the future we might need an array to test on multiple versions OR we will need to write new tests for those versions. Unsure what the solution will need to be, going to make a future work ticket.

Tests/Common/Fixtures/Transaction/Receipt+Fixtures.swift
the-real-adammork: explained on previous comment

Tests/Common/Fixtures/Transaction/Transaction+Fixtures.swift
the-real-adammork: explained on previous comment

Tests/Common/Fixtures/Transaction/Transaction+Fixtures.swift
the-real-adammork: explained on previous comment

Sources/Ledger/BlockVersion.swift
the-real-adammork: https://app.asana.com/0/1200318482173758/1202227741761295/f

Sources/Account/Account+TransactionOperations.swift
carybakker: Ah, so I'm not referring to the rename...  I'm referring to the addition of the block check, where the block check will actually refresh the value, but the fee check pulls the cached value.   I could have put the note further down, but I wanted to associate it with the location where the fee was being pulled.   Really it's regarding the order of block versus fee.

the-real-adammork: I think I understand your comment now, so if the cache is stale the feeStrategy() call will update it, and then the subsequent blockVersion call will essentially always pull from the cache cus it just got updated by the feeStrategy call. I was thinking to make a call to combine both into one function to avoid the double scope nesting but didn't want to add more overhead to this change and figured the extra scope was an ok tradeoff for the time saved implementing.

PR #135, opened 2022-05-13
[iOS] Fix `ConnectionIntTests` & `UrlLoadBalancerIntTests`

carybakker:
### Motivation

ConnectionIntTests and UrlLoadBalancerIntTests began failing at some point, now they're fixed.


Tests/Integration/Network/Protocol/ConnectionIntTests.swift
```diff
@@ -6,14 +6,32 @@ import LibMobileCoin
 @testable import MobileCoin
 import XCTest
 
+// swiftlint:disable line_length
```
the-real-adammork: lets fix the line_length lint warnings

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
```diff
@@ -6,56 +6,133 @@ import LibMobileCoin
 @testable import MobileCoin
 import XCTest
 
+// swiftlint:disable line_length
```
the-real-adammork: lets fix these

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
```diff
@@ -69,20 +146,44 @@ class UrlLoadBalancerIntTests: XCTestCase {
             }
             expectFailure.fulfill()
         }
-        waitForExpectations(timeout: expectationTimeout)
+        waitForExpectations(timeout: timeToWait)
```
the-real-adammork: Because the timeout can be generated/computed from the existing function parameters, lets simplify the call by removing the `andWaitFor timeToWait: Double` and using an extension on `TransportProtocol` something like below:

```swift
extension TransportProtocol {
    static var grpcTimeout: Double = {
        Double(GrpcChannelManager.Defaults.callOptionsTimeLimit.timeout!.nanoseconds) / 1.0e9
    }()
    
    static var httpTimeout: Double = {
        DefaultHttpRequester.defaultConfiguration.timeoutIntervalForRequest
    }()
    
    var timeoutInSeconds: Double {
        // pad timeout of test by 2 seconds over the protocol timeout
        let padTime = 2.0

        switch self.option {
        case .grpc:
            return padTime + Self.grpcTimeout
        case .http:
            return padTime + Self.httpTimeout
        }
    }
}
```

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
```diff
@@ -6,56 +6,133 @@ import LibMobileCoin
...
 
-    let expectationTimeout = 60.0
-    
+    func timeoutInSeconds(for transportProtocol: TransportProtocol) -> Double {
+        // pad timeout of test by 2 seconds over the protocol timeout
+        let padTime = 2.0
+
+        switch transportProtocol.option {
+        case .grpc:
+            // forced unwrap of timeout here in test is ok b/c we want fast failure if not set
```
the-real-adammork: Im ok with force-unwraps in test code, the comment isn't needed.

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
the-real-adammork: then this becomes

```suggestion
        waitForExpectations(timeout: transportProtocol.timeoutInSeconds)
```

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
```diff
@@ -97,20 +198,45 @@ class UrlLoadBalancerIntTests: XCTestCase {
...
+        waitForExpectations(timeout: timeToWait)
+
+        // verify that rotate has been called by checking curIdx of the loadBalancer
+        XCTAssertEqual( 1, consensusUrlLoadBalancerMock.curIdx, "Consensus URL load balancer should have rotated from index 0 to 1")
+    }
+
+    func testConsensusDoesNotRotateAwayFromGoodUrl(using transportProtocol: TransportProtocol, andWaitFor timeToWait: Double) throws {
+        print("\n-----\nTesting Consensus url rotation does not occur with good URL for protocol: \(transportProtocol.description)")
+
+        let fogUrlLoadBalancer = try createValidFogUrlLoadBalancer()
```
the-real-adammork: I think all of these test-configuration lines should be moved into a "fixtures" file. There a bunch of examples if you search the repo, but the idea is to make an internal, named struct for each config permutation thats needed in testing, then you instantiate that struct and use the preconfigured bits in the test. It simplifies the test functions and makes things more readable.

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
```diff
@@ -285,77 +521,90 @@ extension UrlLoadBalancerIntTests {
...
     }
 
-    func createFogViewConnection(transportProtocol: TransportProtocol) throws -> FogViewConnection {
-        let fogUrlLoadBalancer = try createTestFogUrlLoadBalancer()
-
-        // disable url rotation during creation of connection/svc as multipe calls
-        // to nextUrl() will cause rotation away from the initial invalid URL
-        fogUrlLoadBalancer.rotationEnabled = false
-
+    func createConfiguredFogViewConnection
```
the-real-adammork: The fixtures file is a good home for these shared convenience functions

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
carybakker: Updated - I like it.

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 defaults:
-  default-xcode-version: &default-xcode-version "12.2.0"
+  default-xcode-version: &default-xcode-version "13.3.1"
```
the-real-adammork: Is this change still necc. ? If not lets revert this to 12 and make a future work ticket to **add** xcode 13 support for CICD. which means running some things twice once for each version.

Tests/Integration/IntegrationTestFixtures+UrlLoadBalancing.swift
```diff
@@ -0,0 +1,267 @@
...
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+@testable import MobileCoin
+import XCTest
+
+extension IntegrationTestFixtures {
+    static func createValidConsensusUrlLoadBalancer()
```
the-real-adammork: Im know Im nit-picking but this isn't really following the Fixture pattern being used elsewhere in the code-base.

Tests/Integration/IntegrationTestFixtures+UrlLoadBalancing.swift
```diff
@@ -0,0 +1,267 @@
...
+    throws -> SequentialUrlLoadBalancer<FogUrl> {
+        let urlStrings = [
+            IntegrationTestFixtures.invalidFogUrl,
+            IntegrationTestFixtures.network.fogUrl,
+        ]
+        let fogUrls = try FogUrl.make(strings: urlStrings).get()
+        return SequentialUrlLoadBalancer(urls: fogUrls)
+    }
+
+    static func createFogViewConnection
```
the-real-adammork:
```swift
    static func createFogViewConnection(
        for transportProtocol: TransportProtocol,
        using fogUrlLoadBalancer: SequentialUrlLoadBalancer<FogUrl>
    ) throws -> FogViewConnection {
```

review the syntax techniques used in other files for long lines and match that when transforming these

Tests/Integration/Network/Protocol/ConnectionIntTests.swift
```diff
@@ -7,85 +7,66 @@ import LibMobileCoin
...
 
-    func testCallTimeoutWithInvalidUrl(transportProtocol: TransportProtocol) throws {
+    func testInvalidUrlTimeout(using transportProtocol: TransportProtocol) throws {
         let expect = expectation(description: "Making Fog View enclave request")
 
         var request = FogLedger_BlockRequest()
         request.rangeValues = [1..<2]
-        try createFogBlockConnectionWithInvalidUrls(transportProtocol: transportProtocol).getBlocks(request: request) {
-            guard let error = $0.failureOrFulfill(expectation: expect) else { return }
+        try IntegrationTestFixtures.createFogBlockConnectionWithInvalidUrls(
```
the-real-adammork, line 24: fixture pattern is not being properly implemented, can you look at some of the other test files in the project ? The new ones Ive added for memos come to mind, but alot of the other ones use this pattern. If you want to pair up on one of these I could do that.

Tests/Integration/IntegrationTestFixtures+UrlLoadBalancing.swift
the-real-adammork: do some research on other test files, look at 10 or so and you'll get the idea of what fixtures are and how to set them up. They solve for a few things:

- line length, asking for a variable (thats already configured) with `fixture.variable` is alot shorter.
- re-use
- readability

carybakker: Gotcha...  I based all this code on the existing IntegrationTestFixtures.swift file code that was already here, which relied on the 'create' methods as opposed to a more "Fixture" based paradigm...  Will refactor.

Tests/Integration/Network/Protocol/ConnectionIntTests.swift
carybakker: Ya, this code was based on the existing code in IntegrationTestFixtures which was 'create' function based, as opposed to actual Fixtures, per se... will refactor.

Tests/Integration/IntegrationTestFixtures+UrlLoadBalancing.swift
carybakker: Ya, the code, as is, where the open parens is on a newline, was due to the linter warning `Multline Parameters Brackets Violation: Multiline parameters should have their surrounding brackets in a new line. (multiline_parameters_brackets)`.  

It's a little bit misleading, because it'll give you that warning for both the open parens and the close parens if you, say, were to have the first param as a continuation of the function name line...  

So when going through this, while I thought it looked strange, I was trying to follow the `multiline_parameters_brackets` instructions, which literally stated the surrounding brackets should be in a new line for both open parens and close.

Wrapping the first param does get rid of the linter warning for open parens, so it's strange / buggy that it's flagged :/

Tests/Integration/Network/Protocol/ConnectionIntTests.swift
carybakker: Fixed

Tests/Integration/Network/UrlLoadBalancerIntTests/UrlLoadBalancerIntTests.swift
carybakker: fixed

.circleci/config.yml
carybakker: reverted - thanks for catching this

Tests/Integration/IntegrationTestFixtures+UrlLoadBalancing.swift
carybakker: Fixed - seemed like the only issue was the open parens for multi-line parameter lists, where I blindly followed the linter...

Tests/Integration/Fixtures/UrlLoadBalancerFixtures.swift
```diff
@@ -1,53 +1,357 @@
...
-        validUrlsConsensusUrlBalancer = try Self.createValidConsensusUrlLoadBalancer()
         validUrlsFogUrlBalancer = try Self.createValidFogUrlLoadBalancer()
+        validUrlsConsensusUrlBalancer = try Self.createValidConsensusUrlLoadBalancer()
         initialInvalidUrlFogUrlBalancer = try Self.createFogUrlLoadBalancerWithInitialInvalidUrl()
         initialInvalidConsensusUrlBalancer =
             try Self.createConsensusUrlLoadBalancerWithInitialInvalidUrl()
     }
 }
 
+protocol ServiceFixture {
```
the-real-adammork, line 25: SMART

the-real-adammork (approved):
looks great !

PR #139, opened 2022-05-24
Support BlockVersion == 2, TokenId's, Multiple Balances, TxOutContext for constructed Transactions

the-real-adammork:
Soundtrack of this PR: [Zapp - Be Alright](https://www.youtube.com/watch?v=UvZAt7VnJls)

### Motivation

BlockVersion 2 of `mobilecoin` has support for multiple Tokens, this affects: amounts, fees, and how we construct transactions.

### In this PR
* Add `TokenId` object 
* Support multiple `Balance`s
* Fee is now denominated with a Token Id
* Fees now come from a minimumFees map
* TxOutContext with more information for constructed transactions
* Deprecate older public APIs that do not support Tokens

### Future Work
* Fix Linting Warnings


Sources/Account/Balances.swift
```diff
@@ -0,0 +1,39 @@
...
+                tokenId: .MOB)
+        }
+        return balance
+    }
+    
+    let blockCount: UInt64
+    
+    init(balances: [Balance], blockCount: UInt64) {
+        self.balances = balances.reduce(into: [TokenId:Balance](), { result, balance in
+            result[balance.tokenId] = balance
```
carybakker: Do you think it might be worth adding some error checking here to make sure we don't have input with more than one value for a particular token ID in the input array?

e.g. if we have two Balance values in the input balances, both for .Mob, then the second one will overwrite the first...

Just thinking it's a chance to add a check for bad input and prevent the propagation of a hard-to-find error that may pop up in other code that's calling this.

Sources/Transaction/Amount.swift
```diff
@@ -0,0 +1,34 @@
...
+extension Amount {
+    public init(value: UInt64, token: TokenId) {
+        self.value = value
+        self.tokenId = token
+    }
+
+    public init(_ value: UInt64, _ tokenId: TokenId) {
+        self.value = value
+        self.tokenId = tokenId
+    }
```
carybakker: why do we need both of these?

the-real-adammork: the second one has a shorter function signature which im using in a few places to keep the lines more readable.

Sources/Account/Balances.swift
the-real-adammork: ya I guess thats a good point, I could move this reduce out of the init and make the init accept the map instead ?

Sources/Transaction/Amount.swift
carybakker: Sure...  But if the second one is sufficient, wondering why we need the first?   It's not a huge deal to have both, but a minor nit would be to make the parameter names match for tokenId.

Sources/Account/Balances.swift
carybakker: Sure - at least that way we could verify that we aren't vulnerable to bad data as input for the class b/c the map can only have one value per token ID by definition...

Sources/Transaction/Fee/BlockchainMetaFetcher.swift
```diff
@@ -116,7 +111,13 @@ final class BlockchainMetaFetcher {
...
                 let blockVersion = BlockVersion(response.networkBlockVersion)
-                self.cacheMeta(minimumFee: minimumFee, blockVersion: blockVersion) {
+                let minimumFees = response.minimumFees.reduce(into: [TokenId: UInt64](), {
+                    $0[TokenId($1.key)] = $1.value
+                })
+                self.cacheMeta(
+                    minimumFee: minimumFee,
+                    minimumFees: minimumFees,
+                    blockVersion: blockVersion) {
                     completion(.success($0))
```
carybakker, line 118: We could drop minimumFee here and just use minimumFees everywhere else, if we just check for the presence of .MOB in the response and then set to the DEFAULT_MINIMUM_FEE if a value for .MOB is not present. 

i.e. just set the value for .MOB in the fees if it doesn't exist...

Then we wouldn't need to do that check in other places, we'd just rely on minimumFees[tokenId], and that would simplify the getOrFetch code in lines 68-69, right? 

Also, for the single minimum fee value, the logic will set to a minimum if the value is zero, but if we receive `minimumFees` in the response, we do not override a value of zero with DEFAULT_MINIMUM_FEE like we do for the single fee version of the code.   Should they be the same with respect to the default value logic?

Sources/Transaction/Fee/BlockchainMetaFetcher.swift
```diff
@@ -36,66 +36,61 @@ final class BlockchainMetaFetcher {
...
         fetchCache {
-            if let minimumFee = $0?.minimumFee {
+            if let minimumFee = $0?.minimumFees[tokenId] ??
+               (tokenId == .MOB ? $0?.minimumFee : nil) {
                 completion(.success(minimumFee))
             } else {
                 self.fetchMeta {
                     completion($0.map { cache in
-                        cache.minimumFee
+                        cache.minimumFees[tokenId] ?? cache.minimumFee
```
carybakker: Shouldn't we throw an error here?  I can see using the minimumFee for .MOB, but using the .MOB minimum fee value for a tokenId that's not found would be incorrect, right?

Sources/Account/Balance.swift
```diff
@@ -5,95 +5,146 @@
...
 
-        let (amountLowMobInt, amountLowPicoFrac) = { () -> (UInt32, UInt64) in
-            // 10^12 = 1_000_000_000_000
-            let mobParts = amountPicoMobLow.quotientAndRemainder(dividingBy: 1_000_000_000_000)
-            return (UInt32(mobParts.quotient), mobParts.remainder)
+        
+        let significantDigits = tokenId.significantDigits
+        
+        let divideBy = UInt64(pow(Double(10), Double(significantDigits)))
+        let (amountLowMobInt, amountLowFrac) = { () -> (UInt64, UInt64) in
```
carybakker: we should change all the amount*Mob* variable names to just amount* (like `amountLowMobInt` -> `amountLowInt`.

Sources/Account/Balance.swift
```diff
@@ -5,95 +5,146 @@
...
+            return (UInt64(mobParts.quotient), mobParts.remainder)
         }()
 
-        let (amountHighMobInt, amountHighPicoFrac) = { () -> (UInt32, UInt64) in
-            // Intermediary = base of 5^-12 MOB
-            let amountHighIntermediary = UInt64(amountPicoMobHigh) << 52
-            // 5^12 = 244_140_625
-            let mobParts = amountHighIntermediary.quotientAndRemainder(dividingBy: 244_140_625)
-            return (UInt32(mobParts.quotient), mobParts.remainder << 12)
+        let (amountHighMobInt, amountHighFrac) = { () -> (UInt64, UInt64) in
```
carybakker: amountHighMobInt -> amountHighInt

Sources/Account/Balance.swift
```diff
@@ -5,95 +5,146 @@
...
         }()
 
-        let amountPicoFracParts = (amountLowPicoFrac + amountHighPicoFrac).quotientAndRemainder(
-            dividingBy: 1_000_000_000_000)
+        let amountFracParts = (amountLowFrac + amountHighFrac).quotientAndRemainder(
+            dividingBy: divideBy)
 
-        let amountMobInt = amountLowMobInt + amountHighMobInt + UInt32(amountPicoFracParts.quotient)
-        let amountPicoFrac = amountPicoFracParts.remainder
+        let amountMobInt = amountLowMobInt + amountHighMobInt + UInt64(amountFracParts.quotient)
```
carybakker: same... remove Mob from variable names

Sources/Account/Balance.swift
```diff
@@ -5,95 +5,146 @@
...
         // amountPicoFrac = (amountLowPicoFrac + amountHighPicoFrac) % 10^12
 
-        let (amountLowMobInt, amountLowPicoFrac) = { () -> (UInt32, UInt64) in
-            // 10^12 = 1_000_000_000_000
-            let mobParts = amountPicoMobLow.quotientAndRemainder(dividingBy: 1_000_000_000_000)
-            return (UInt32(mobParts.quotient), mobParts.remainder)
+        
+        let significantDigits = tokenId.significantDigits
+        
+        let divideBy = UInt64(pow(Double(10), Double(significantDigits)))
```
carybakker, line 128: I feel like we should be using a precomputed value instead of using pow for every balance... since the divideBy could be calculated for every tokenId once instead of every time a Balance is created

Sources/Account/Balance.swift
```diff
@@ -5,95 +5,146 @@
...
 
-        let (amountHighMobInt, amountHighPicoFrac) = { () -> (UInt32, UInt64) in
-            // Intermediary = base of 5^-12 MOB
-            let amountHighIntermediary = UInt64(amountPicoMobHigh) << 52
-            // 5^12 = 244_140_625
-            let mobParts = amountHighIntermediary.quotientAndRemainder(dividingBy: 244_140_625)
-            return (UInt32(mobParts.quotient), mobParts.remainder << 12)
+        let (amountHighMobInt, amountHighFrac) = { () -> (UInt64, UInt64) in
+            let amountHighIntermediary = UInt64(amountHigh) << (64 - significantDigits)
+            let factored = UInt64(pow(Double(5), Double(significantDigits)))
```
carybakker, line 136: thoughts on how we could compute factored exactly once per tokenId (same for divideBy)?

Sources/Account/Account+TransactionOperations.swift
```diff
@@ -56,7 +57,7 @@ extension Account {
             }
 
             let (unspentTxOuts, ledgerBlockCount) =
-                account.readSync { ($0.unspentTxOuts, $0.knowableBlockCount) }
+            account.readSync { ($0.unspentTxOuts(tokenId: amount.tokenId), $0.knowableBlockCount) }
```
carybakker, line 60: just an observation...

unspentTxOuts calls unspentTxOutsAndBlockCount (ref: Account.swift), which internally calls self.knowableBlockCount, which has to check `for txOut in allTxOutTrackers` to calculate the computed property value.

The separate call to get knowableBlockCount will re-do the calculation of the computed property for knowableBlockCount.

Not sure how impactful that calculation is, but since there's a call to get both from account in one call, maybe we should use that instead.

Sources/Account/Account.swift
```diff
@@ -46,12 +46,28 @@ final class Account {
...
+    var cachedTxOutTokenIds: Set<TokenId> {
+        Set(allTxOutTrackers
+            .map { $0.knownTxOut.tokenId })
+    }
+
+    func cachedBalance(for tokenId: TokenId) -> Balance {
         let blockCount = knowableBlockCount
         let txOutValues = allTxOutTrackers
             .filter { $0.receivedAndUnspent(asOfBlockCount: blockCount) }
+            .filter { $0.knownTxOut.tokenId == tokenId }
```
carybakker, line 63: This can be done in one pass with && on the two criteria instead of two passes..

Whether or not it matters depends on how many items there might be, and how often cachedBalance might be referenced...

I like the way the chained .filter calls look in the code, just want to call out that it can be a little bit more efficient (if it matters)

Sources/Account/Account.swift
```diff
@@ -72,14 +88,17 @@ final class Account {
...
     }
 
-    var unspentTxOutsAndBlockCount: (txOuts: [KnownTxOut], blockCount: UInt64) {
+    func unspentTxOutsAndBlockCount(
+        tokenId: TokenId
+    ) -> (txOuts: [KnownTxOut], blockCount: UInt64) {
         let knowableBlockCount = self.knowableBlockCount
         let txOuts = allTxOutTrackers
             .filter { $0.receivedAndUnspent(asOfBlockCount: knowableBlockCount) }
+            .filter { $0.knownTxOut.tokenId == tokenId }
```
carybakker, line 106: This can be done in one filter pass with && instead of having to do two passes, but I like the way the code looks... 

Whether or not it matters kind of depends on how large the initial array (and filtered array) might be, and how often this is called.

Sources/Account/Account.swift
```diff
@@ -46,12 +46,28 @@ final class Account {
...
         let txOutValues = allTxOutTrackers
             .filter { $0.receivedAndUnspent(asOfBlockCount: blockCount) }
+            .filter { $0.knownTxOut.tokenId == tokenId }
             .map { $0.knownTxOut.value }
-        return Balance(values: txOutValues, blockCount: blockCount)
+        return Balance(values: txOutValues, blockCount: blockCount, tokenId: tokenId)
+    }
+
+    var cachedBalances: Balances {
+        let balances = cachedTxOutTokenIds.map { cachedBalance(for: $0) }
```
carybakker: I like the elegance of this, but it's also going to scan all the txOuts completely just to get the set of tokenIDs, and then per tokenID will do full passes over the txOuts again to check for received and unspent, and another for the tokenID check...

It can all be done with one pass over the txOuts... 

Right now not really a big deal, but if we have a lot of tokenIDs and if there is a lot of activity, could this be a potential performance problem?

Sources/Ledger/PartialTxOut.swift
```diff
@@ -22,23 +23,29 @@ extension PartialTxOut {
...
 extension PartialTxOut {
     init?(_ txOut: External_TxOut) {
-        guard let commitment = Data32(txOut.amount.commitment.data),
+        guard let commitment = Data32(txOut.maskedAmount.commitment.data),
               let targetKey = RistrettoPublic(txOut.targetKey.data),
               let publicKey = RistrettoPublic(txOut.publicKey.data)
         else {
             return nil
         }
+        guard [0,4,8].contains(txOut.maskedAmount.maskedTokenID.count) else {
```
carybakker: why not just add this to the long list of prior guard statements?

Sources/Ledger/PartialTxOut.swift
```diff
@@ -48,17 +55,22 @@ extension PartialTxOut {
...
                                                     maskedValue: txOutRecord.txOutAmountMaskedValue,
+                                                    maskedTokenId: txOutRecord.txOutAmountMaskedTokenID,
                                                     publicKey: publicKey,
                                                     viewPrivateKey: viewKey),
               Self.isCrc32Matching(commitment, txOutRecord: txOutRecord)
         else {
             return nil
         }
-
+        guard [0,4,8].contains(txOutRecord.txOutAmountMaskedTokenID.count) else {
```
carybakker: why not just add this to the long list of prior guard statements?  (like the Self.isCrc32Matching check)?

Sources/Ledger/PartialTxOut.swift
```diff
@@ -80,3 +92,31 @@ extension PartialTxOut {
...
+        case 4:
+            guard let value = self.to(type: UInt32.self) else { return nil }
+            return UInt64(value)
+        case 8:
+            return self.to(type: UInt64.self)
+        default:
+            return nil
+        }
+    }
+}
```
carybakker: Maybe move to the `MobileCoin/Core/Utils/Data/Data+TypeConversions.swift` ?

Sources/LibMobileCoin/McConstants.swift
```diff
@@ -74,7 +74,13 @@ extension McConstants {
     static let DEFAULT_SUBADDRESS_INDEX: UInt64 = 0
 
     /// An account's "default change address" is its first subaddress.
-    static let DEFAULT_CHANGE_SUBADDRESS_INDEX: UInt64 = 1
+    static let DEFAULT_CHANGE_SUBADDRESS_INDEX = UInt64.max - 1
```
carybakker, line 77: Just wondering why we do max-1 instead of just max?  Is that in a spec somewhere?

Tests/Common/Fixtures/Account/BalanceTests+Fixtures.swift
```diff
@@ -0,0 +1,6 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+@testable import MobileCoin
```
carybakker: Delete if not needed

Tests/Common/Fixtures/Fog/FogBlockConnection+Fixtures.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import LibMobileCoin
+@testable import MobileCoin
+import XCTest
+
```
carybakker: delete file if unused

Tests/Integration/IntegrationTestFixtures.swift
```diff
@@ -8,7 +8,7 @@
 import XCTest
 
 enum IntegrationTestFixtures {
-    static let network: NetworkPreset = .testNet
```
carybakker, line 11: Does it matter where this defaults?  not right now, right?

Sources/Transaction/Amount.swift
carybakker: What do you think of:
`public init(_ value: UInt64, in tokenId: TokenId)`  

so the client code will look like:

` Amount(value, in: .MOB)` 

or

`Amount(value, in: tokenId)`?

Sources/Ledger/TxOutUtils.swift
```diff
@@ -59,40 +36,92 @@ enum TxOutUtils {
...
-                        // supplied incorrectly.
-                        logger.warning("error: \(redacting: error)")
-                        return nil
-                    default:
-                        // Safety: mc_tx_out_reconstruct_commitment should not throw
-                        // non-documented errors.
-                        logger.warning("Unhandled LibMobileCoin error: \(redacting: error)")
-                        return nil
+        reconstructCommitment(maskedValue: maskedValue,
+                              maskedTokenId: Data(),
```
carybakker: Does this work?  I didn't drill all the way down, but does passing an empty Data in end up defaulting to .MOB or something when attempting to pull out the tokenId from maskedTokenId?

Sources/Transaction/TransactionBuilder.swift
```diff
@@ -294,7 +317,13 @@ final class TransactionBuilder {
             fogResolver.withUnsafeOpaquePointer { fogResolverPtr in
                 // Safety: mc_transaction_builder_create should never return nil.
                 withMcInfallible {
-                    mc_transaction_builder_create(fee, tombstoneBlockIndex, fogResolverPtr, memoBuilderPtr, blockVersion)
+                    mc_transaction_builder_create(
+                            fee.value,
+                            tokenId.value,
```
carybakker: Seems strange to discard the tokenId of the Amount parameter (`fee`), but have a separate tokenId param for the builder.

It also would seem strange to use the tokenId of the fee to infer the tokenId for the transaction builder.

Maybe this function should also only take a value param instead of an Amount, so that we're not receiving information that is not used...

Or, we can leave it like it is, and check that the tokenId of Amount matches the tokenId param, but we'd have to throw an exception or handle that error condition somehow.

carybakker (changes requested):
Had a few comments / suggestions / questions...

Tests/Unit/Account/AccountTests.swift
```diff
@@ -14,8 +14,9 @@ class AccountTests: XCTestCase {
         let balanceUpdater = try Self.balanceUpdater(fogService: testFogService)
 
         let expect = expectation(description: "Checking account balance")
-        balanceUpdater.updateBalance {
-            guard let balance = $0.successOrFulfill(expectation: expect) else { return }
+        balanceUpdater.updateBalances {
```
carybakker, line 17: I feel like we should something like what's done for the http / grpc protocols... where we iterate over the supported token IDs for each test.

wdyt?

Sources/Account/Balance.swift
the-real-adammork: I think this is a premature optimization, if there is evidence that these calculation are causing noticeable slowdowns, id be willing to change my mind.

Sources/Account/Balance.swift
the-real-adammork: I think this is a premature optimization, if there is evidence that these calculation are causing noticeable slowdowns, id be willing to change my mind.

Sources/Account/Account+TransactionOperations.swift
the-real-adammork: think this is premature optimization as-well, bernie tried to optimize these loops and ran into some new bugs. Im happy to optimize in the future if its needed.

Sources/Account/Account.swift
the-real-adammork: I prefer the readability, if these become bottlenecks we know how to improve but I don't think they are right now.

Sources/Account/Account.swift
the-real-adammork: again, not sure the changes make sense here, happy to be provided evidence that it is significant, in the meantime we could add some new cases to the PerformanceTests to benchmark these functions but Id like to punt that to future work in the interest of the timelines we are on.

Sources/LibMobileCoin/McConstants.swift
the-real-adammork: https://github.com/mobilecoinfoundation/mcips/pull/36

Tests/Integration/IntegrationTestFixtures.swift
the-real-adammork: not really

Tests/Unit/Account/AccountTests.swift
the-real-adammork: Right now and in the near future there will only ever be MOB and MOBUSD so I think its ok to copy paste or be explicit, but in the future, when we have more tokens, and we have a way to fetch that list, it will make sense to be more dynamic and do some looping like you've suggested.

Tests/Common/Fixtures/Fog/FogBlockConnection+Fixtures.swift
the-real-adammork: deleted, ty

Tests/Common/Fixtures/Account/BalanceTests+Fixtures.swift
the-real-adammork: deleted, ty

Sources/Transaction/Fee/BlockchainMetaFetcher.swift
the-real-adammork: ok ya I like your idea, will simplify some of the code paths. Ill look into override going to zero, it should be going
 `DEFAULT_MINIMUM_FEE`

Sources/Transaction/Fee/BlockchainMetaFetcher.swift
the-real-adammork: will investigate and see what android does, im told by blockchain that this will never happen tho.

Sources/Transaction/Amount.swift
the-real-adammork: ok will do this

Sources/Ledger/TxOutUtils.swift
the-real-adammork: empty Data will be interpreted as tokenId .MOB

maybe would be better to define this as a constant somewhere to improve readability

Sources/Ledger/PartialTxOut.swift
the-real-adammork: ok ya good idea

Sources/Ledger/PartialTxOut.swift
the-real-adammork: ya not sure, ill move it up

Sources/Ledger/PartialTxOut.swift
the-real-adammork: ya not sure, ill move it up

the-real-adammork:
@carybakker grouping suggestions into a todo list,

TODO:
- [x] Implement suggestion for changing fee/fee-map behavior
- [x]  Add performance test case for some of the txOut map computations - tokenId, etc.. - verify not a bottleneck
- [x] Amount initializer function signature changes
- [x] Define empty Data() as a constant for maskedTokenId == .MOB (legacy compatability)
- [x] Move data extensions to new home
- [x] Combine [0, 4, 8] guard with other guards.
- [x] make Balances accept a map instead of an array
- [x] Fix variable names in Balance struct to be agnostic
- [x] TransactionBuilder.init() changes

Sources/Account/Account.swift
the-real-adammork: performances tests added, the filtering is not the bottleneck, its the creation of the `Balance` structs, and im not sure optimizing that should be in the scope of this PR especially when we know for the near future there will only be at max 2 `Balance` structs.

the-real-adammork: performance slows down at > 1000 balance structs.

Sources/Account/Account.swift
the-real-adammork: verified not a bottleneck

Sources/Account/Account+TransactionOperations.swift
the-real-adammork: verified not a bottleneck

Sources/Account/Balance.swift
the-real-adammork: creating `Balance` structs gets slow at > 1000, I have performance tests now, we should re-visit this when the number of available tokens exceeds 100, as that may be a point at which it affects user's perceived performance.

Sources/Account/Balance.swift
the-real-adammork: we can revisit when we get to > 100 available tokens

Sources/Account/Balance.swift
the-real-adammork: done

Sources/Account/Balance.swift
the-real-adammork: done

Sources/Account/Balance.swift
the-real-adammork: done

Tests/Unit/Account/AccountTests.swift
the-real-adammork: verified this is not a performance bottleneck

Sources/Transaction/Fee/BlockchainMetaFetcher.swift
the-real-adammork: changed to `cache.minimumFees[tokenId] ?? McConstants.DEFAULT_MINIMUM_FEE` I think this is better behavior but again this is something that should never happen assuming the token is valid.

Sources/Transaction/TransactionBuilder.swift
the-real-adammork: changed the init to use the tokenId from the fee Amount

v1.2.0 does not support mixed transaction so all txOuts and the fee will have the same tokenId

Sources/Transaction/Amount.swift
the-real-adammork: done

the-real-adammork:
@carybakker tests passing

Tests/Common/Fixtures/Account/AccountKey+Fixtures.swift
```diff
@@ -163,6 +188,38 @@ extension AccountKey.Fixtures.Init {
...
+//        """
+    fileprivate static let fogAuthoritySpkiHex = "30820222300d06092a864886f70d01010105000382020f003082020a0282020100c853a8724bc211cf5370ed4dbec8947c5573bed0ec47ae14211454977b41336061f0a040f77dbf529f3a46d8095676ec971b940ab4c9642578760779840a3f9b3b893b2f65006c544e9c16586d33649769b7c1c94552d7efa081a56ad612dec932812676ebec091f2aed69123604f4888a125e04ff85f5a727c286664378581cf34c7ee13eb01cc4faf3308ed3c07a9415f98e5fbfe073e6c357967244e46ba6ebbe391d8154e6e4a1c80524b1a6733eca46e37bfdd62d75816988a79aac6bdb62a06b1237a8ff5e5c848d01bbff684248cf06d92f301623c893eb0fba0f3faee2d197ea57ac428f89d6c000f76d58d5aacc3d70204781aca45bc02b1456b454231d2f2ed4ca6614e5242c7d7af0fe61e9af6ecfa76674ffbc29b858091cbfb4011538f0e894ce45d21d7fac04ba2ff57e9ff6db21e2afd9468ad785c262ec59d4a1a801c5ec2f95fc107dc9cb5f7869d70aa84450b8c350c2fa48bddef20752a1e43676b246c7f59f8f1f4aee43c1a15f36f7a36a9ec708320ea42089991551f2656ec62ea38233946b85616ff182cf17cd227e596329b546ea04d13b053be4cf3338de777b50bc6eca7a6185cf7a5022bc9be3749b1bb43e10ecc88a0c580f2b7373138ee49c7bafd8be6a64048887230480b0c85a045255494e04a9a81646369ce7a10e08da6fae27333ec0c16c8a74d93779a9e055395078d0b07286f9930203010001"
+
```
carybakker, line 220: did you mean to leave this in here?

carybakker (approved):
Awesome work cranking out all the fixes so fast!
Looks great!

Only comment was around a chunk of commented lines for SPKI... if you didn't mean to leave it, should probably delete the commented chunk.

Have a great week!
Cary

PR #144, opened 2022-06-01
Change to filter non-MOB transactions for legacy txOuts property of A…

carybakker:
Quick change to filter txOuts to only MOB so non MOB transactions don't apear in Signal's history.

Note, there may be a better alternate approach for iOS - I'll submit another PR early tomrrow.

I think for iOS, it'll be better to keep the AccountActivity struct as-is, and to modify the MobileCoinClient's methods as follows:

```
    public var accountActivity: AccountActivity {
        accountActivity(for: .MOB)
    }

    public func accountActivity(for tokenId: TokenId) -> AccountActivity {
        accountLock.readSync { $0.cachedAccountActivity(for: tokenId) }
    }
```

and add corresponding support to filter for cachedAccountActivity...

### DIFF
--- a/Sources/Account/AccountActivity.swift
+++ b/Sources/Account/AccountActivity.swift
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//  Copyright (c) 2020-2022 MobileCoin. All rights reserved.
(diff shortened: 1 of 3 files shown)


carybakker:
Closed as we went with PR145 instead

PR #146, opened 2022-06-01
Alternate approach to filtering transactions to only MOB for legacy c…

carybakker:
@the-real-adammork lmk what you think of this approach vs the other

### DIFF
--- a/Sources/Account/Account.swift
+++ b/Sources/Account/Account.swift
@@ -75,12 +75,27 @@ final class Account {
         return Balances(balances: balances, blockCount: knowableBlockCount)
     }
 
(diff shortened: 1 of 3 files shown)


the-real-adammork:
changes so far look good, but I think we should include a new variable `tokenId` on the `AccountActivity` struct. This lets consumers easily get the token we filtered on without looping through the txOuts.

Or we don't add tokenId, but in that case we should add a getter to the public API for consumers to get "all" txOuts for "all" tokenIds.

carybakker:
@the-real-adammork  updated AccountActivity to include tokenId

the-real-adammork (approved):

PR #153, opened 2022-06-14
Update SwiftNIO dependency to 2.40.0

collinjackson:
Soundtrack of this PR: [link to song that really fits the mood of this PR](https://www.youtube.com/watch?v=AX_FxBSEh1A)

### Motivation

SwiftNIO team fixed a CocoaPods link error issue that was preventing clients from building with parallelization enabled: https://github.com/apple/swift-nio/issues/2073#issuecomment-1117138500

This change updates the dependency to pick up the new version so clients can use it.

### In this PR

Update MobileCoin.podspec to SwiftNIO 2.40.0
Regenerate Example/Podfile.lock and ExampleHTTP/Podfile.lock


the-real-adammork:
can you remake your PR with just your commit off of the latest `master` ?

Also, I recently "downgraded"/locked-in an older version of SwiftNIO because of some issues I was seeing, so Id like to verify that this new version doesn't cause a regression on the issue I was fixing for.

This is that commit - https://github.com/mobilecoinofficial/MobileCoin-Swift/commit/af234e621c145ef51029fafb2f48a0fd7d6ae182

this was the PR description

```
### Motivation

Version lock GRPC and SwiftNIO dependencies to preserve network robustness when connections are idle. Integrating apps may see this error after a connection goes idle `NIOCore/EventLoopFuture.swift:429: Precondition failed: leaking an unfulfilled Promise`.

### In this PR
* Version locking for GRPC and SwiftNIO dependencies, latest versions cause errors
* Revert changes that were implemented based on higher-version's deprecations
```

after you rebase/remake the PR I can do some testing.

collinjackson:
Updated, PTAL

collinjackson:
If @the-real-adammork's testing shows that the leak issue is still present, the options are:

1. make a reduced test case and file an issue against SwiftNIO
1. fork the downgraded version and land a targeted fix for the compilation issue
1. punt, live with slow builds, maybe it resolves itself on its own

the-real-adammork (approved):
did several tests in the signal app, no issues found.

PR #155, opened 2022-06-17
RTH High Level APIs

the-real-adammork:
Soundtrack of this PR: [Sade - Smooth Operator](https://www.youtube.com/watch?v=4TYv2PhG89A)

### Motivation

The primitives for Recoverable Transaction History - https://github.com/mobilecoinfoundation/mcips/blob/main/text/0004-recoverable-transaction-history.md - have already been implemented, but making it easy to use for SDK consumers requires some additional work. The high-level API accepts a list of "contacts" (or a single contact) and optionally a list of transactions. It returns a sorted "history" of transactions with memo's recovered if we're able to validate them, and the associated contact, if its a match.

### In this PR
* Public High-Level APIs that return a recovered transaction history
* Contact-like protocol PublicAddressProvider that consumers can use to make contacts
* Unit tests


Sources/MobileCoinClient.swift
```diff
@@ -80,18 +80,34 @@ public final class MobileCoinClient {
...
+    public func recoverTransactions<Contact: PublicAddressProvider>(
+        contacts: Set<Contact>
+    ) -> [HistoricalTransaction] where Contact: Hashable {
+        recoverTransactions(allAccountActivity().txOuts, contacts: contacts)
+    }
 
-        Get a set of all tokenIds that are in TxOuts owned by this account with:
+    public func recoverContactTransactions<Contact: PublicAddressProvider>(
+        contact: Contact
+    ) -> [HistoricalTransaction] where Contact: Hashable {
```
carybakker, line 91: what about using 'for' and dropping the 'Contact' in the name, like this (same for other functions, i.e. prefix the contact and contacts param with 'for') so it reads well, like "recover transactions for contact":
```suggestion
    public func recoverTransactions<Contact: PublicdAdressProvider>(for contact: Contact
    ) -> [HistoricalTransaction] where Contact: Hashable {
```

Sources/Transaction/Memos/SenderWithPaymentRequestMemo.swift
```diff
@@ -4,17 +4,22 @@
...
+extension SenderWithPaymentRequestMemo: Equatable, Hashable { }
+
 struct RecoverableSenderWithPaymentRequestMemo {
     let memoData: Data64
     let addressHash: AddressHash
-    private let accountKey: AccountKey
     let txOutPublicKey: RistrettoPublic
+    private let accountKey: AccountKey
 
     init(_ memoData: Data64, accountKey: AccountKey, txOutPublicKey: RistrettoPublic) {
```
carybakker, line 24: do you want to add the '64' suffix here for consistency? (and above on line 19 as well?)  I noticed you updated the others...  Or there's likely a paradigm for when you use the 64 or not in the variable name or not...is there a pattern you're following for when to use / not use (presuming the type is actually Data64 and not just Data)?

carybakker (approved):

Sources/MobileCoinClient.swift
the-real-adammork: This type of naming was common in objective-C but I think it was dropped in swift4 or 5. So generally the parameter name should be semantic even for the first parameter. 

So I think its best to keep the current naming

Sources/Transaction/Memos/SenderWithPaymentRequestMemo.swift
the-real-adammork: The 64 suffix was added because the type of the memoData was 64 bytes in length. Wouldn't make sense to add that to the account key.

PR #160, opened 2022-07-12
return consensus block index from submitTransaction

stevenspiel:
### Motivation

[task](https://www.pivotaltracker.com/story/show/182613898)

[comment on Moby Idempotence document](https://docs.google.com/document/d/1ayX6H_EjeCRVhu8lqUi7y1-D14RqyGDrewu70XcpYGY/edit?disco=AAAAbNCWhqI):

> Alex Voloshyn: 
> 1. KeyImages can be retrieved from TxOuts. (i.e. OwnedTxOut.getKeyImage)
> 2. Transaction object contains all KeyImages as a set (i.e. transaction.getKeyImages)
> 3. Fog View, Fog Ledger, and Consensus are all separate services with separate block heights. The SDK simplifies working with Fog by synchronizing View and Ledger block heights, but not consensus. So if you see that consensus is at block 1000 and it returns ContansSpentImages you will need to query Fog until it reports block height 1000 as well. (getTransactionStatus returns Status that contains current Fog block index, AccountSnapshot contains block index, etc). **SubmitTransaction does not return the consensus block index, so we will need to either add that or make BlockchainClient block height public to have it available at all times, not only after you submit**.

### In this PR
* return consensus block index from `submitTransaction`

### Related PR (Android):

https://github.com/mobilecoinofficial/android-sdk/pull/94

### DIFF
--- a/Sources/Common/Errors.swift
+++ b/Sources/Common/Errors.swift
@@ -142,13 +142,25 @@ extension DefragTransactionPreparationError: CustomStringConvertible {
     }
 }
 
+public struct SubmitTransactionError: Error {
+    public let submissionError: TransactionSubmissionError
+    public let consensusBlockCount: UInt64?
+}
+
+extension SubmitTransactionError: CustomStringConvertible {
+    public var description: String {
+        "Submit Transaction Error: Consensus Block Count == \(consensusBlockCount), " +
+        "\(submissionError)"
+    }
+}
+
 public enum TransactionSubmissionError: Error {
     case connectionError(ConnectionError)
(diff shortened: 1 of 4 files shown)


stevenspiel:
@the-real-adammork How do you want to deprecate the old behavior?

the-real-adammork:
Updated with a new Error type that wraps the existing one and the consensus block count (when avail). Deprecated the old call.

stevenspiel:
Fantastic! Looks great!

the-real-adammork (approved):

PR #164, opened 2022-07-14
Make mobBalance public to avoid needing to deal with optional

the-real-adammork:
Soundtrack of this PR: [DeBarge- Stay With Me](https://www.youtube.com/watch?v=CinavBPMy-I)

### Motivation

SDK consumers that want to upgrade to the new public APIs will run into an optional `Balance` where this was previously non-optional. We can publicly expose the non-optional fallback that was previously scoped internal.

### In this PR
* Publicly expose `mobBalance` to avoid optional


carybakker (approved):
👍

PR #165, opened 2022-07-14
Add a GitHub Action to run iOS SDK tests

carybakker:
Soundtrack of this PR: [The Chemical Brothers - The Test](https://www.youtube.com/watch?v=yhS9LnDoo_w)

### Motivation

Integration tests were being run manually, requiring manual changes to the code for environment configuration, as well as having to run the tests on local developer machines with no official record of the test being archived for reference / verification by others.

### In this PR
* Pivotal Tracker Task: [[iOS] Add a GitHub Action to run iOS SDK tests](https://www.pivotaltracker.com/story/show/182376333)
* Support is available to automatically run tests against:
  * mobileDev
  * testNet
  * Alpha

### Future Work
* Make it possible to run against dynamically configured environments
* Generate test account wallets from seed value (ref: zoom meeting between @the-real-adammork @carybakker @jgreat on 2022-07-14) and use assigned range of test accounts (work on this to be continued)


scripts/setkeysfor
```diff
@@ -0,0 +1,38 @@
...
+    echo "Naming substitution complete"
+    echo "----------------------------"
+  else
+    echo "----------------------------"
+    echo "Keys files generated with correct name."
+    echo "----------------------------"
+  fi
+}
+
+generate_and_rename_keys $1
```
the-real-adammork, line 32: I sorta liked the old way this was setup where the script ran this function twice for each project, that means we only have to maintain one script. 

Also, the old version was sourcing a config file with the environment variables, are these now set by the GHA ?

the-real-adammork (approved):
looks great, fix the failing CICD, then merge

PR #168, opened 2022-07-22
Add encrypted secrets file to repo, automate initialization of cocoapod-keys for project setup

the-real-adammork:
Soundtrack of this PR: [Mala - Alicia](https://www.youtube.com/watch?v=zpV7radKuwo)

## Motivation

This repo needs some "secrets" for the unit & integration tests to run. In the past we've been passing them around as tribal knowledge, but thats messy and error-prone. This PR addresses those problems by using asymmetric cryptography. 

All contributors to this repo will create a public/private keypair which will be stored in the macOS Keychain (`security`). New contributors will do this by running `scripts/generate_and_save_contributor_keys`. It will add the new public key to a file: `secrets/contributor_public_keys`. Then, they will check in that change, and ask an existing contributor to "re-encrypt" the secrets file with this additional public key. The "re-encryption" process creates an encrypted file that can be decrypted by any of the contributors (with their matching private key).

> Most importantly, how do we use these secrets ? 

They're used by `cocoapod-keys` to "generate" `MobileCoinKeys.h` `MobileCoinKeys.m` files which we pull from at runtime for integration & unit testing. Assuming a contributor has already gone through "New Contributor" steps and the secrets have been re-encrypted, they can re-initialize the project with:

```
$ scripts/reinitialize_keys
```

### Notes

Im using this tool [$ age](https://github.com/FiloSottile/age) to do the encryption and decryption. The scripts will attempt to install it with brew (if necc.)

#### New Contributor

The new contributor checks out the repo and runs:

```bash
$ scripts/generate_and_save_contributor_keys
...
$ git add secrets/contributor_public_keys
...
$ git commit -m 'new contributor public key added, please re-encrypt the secrets for me'
```

Then an existing contributor checks out this change and runs:

```bash
$ scripts/reencrypt_secrets
...
$ git add secrets/keys.encrypted
...
$ git commit -m 'keys re-encrypted for new contributor'
```

#### Initial encryption workflow

To encrypt a local secrets file with all the contributor public keys, run:

```bash
$ scripts/encrypt_secrets_file ~/Desktop/keys.decrypted
...
$ git add secrets/keys.encrypted
...
$ git commit -m 'adding completely new secrets file, encrypted with all the contributor public keys'
```

#### Decrypt the secrets

To decrypt the secrets, and print them to `STDOUT`

```bash
$ scripts/decrypt_secrets
```

#### Re-initialize the project 

> Assume you've gone through new contributor flow

Run the following, which will do all the steps necc to decrypt the secrets, regenerate the files we use at runtime with `cocoapod-keys`, and then re-build the xcode project with `make lock`.

```bash
$ scripts/reinitialize_keys; make lock;
```

### In this PR
* All scripts needed to add contributors, encrypt initial secrets file, decrypt secrets, and re-encrypt secrets
* `scripts/reinitialize_keys` will decrypt the secrets and re-generate the "cocoapod-keys" files used in the project, renaming if necc.
* Modularize the code in `setkeysfor` so its shared with the other scripts. This is currently used by the GHA, so should be tested.

### Future Work
* Incorporate these workflows into the Makefile and other places where it could be useful.
* Update README.md with these steps


carybakker (commented):
Log of attempt to run: 
```
carybakker@Carys-MBP MobileCoin-Swift % ./scripts/generate_and_save_contributor_keys 
----------------------------
Checking dependencies
----------------------------
age is not installed. installing with `brew install age`
==> Downloading https://ghcr.io/v2/homebrew/core/age/manifests/1.0.0
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/age/blobs/sha256:57281390b7c051f73ed13e7e35413e42eb9bd918812bc99de38ef3817ac46dee
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:57281390b7c051f73ed13e7e35413e42eb9bd918812bc99de38ef381
######################################################################## 100.0%
==> Pouring age--1.0.0.arm64_monterey.bottle.tar.gz
🍺  /opt/homebrew/Cellar/age/1.0.0: 8 files, 6MB
==> Running `brew cleanup age`...
Disable this behaviour by setting HOMEBREW_NO_INSTALL_CLEANUP.
Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).
all dependencies installed
carybakker@Carys-MBP MobileCoin-Swift % git status                             
On branch task/age-secrets-manager
Your branch is up to date with 'origin/task/age-secrets-manager'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	1

nothing added to commit but untracked files present (use "git add" to track)
carybakker@Carys-MBP MobileCoin-Swift % cat 1
security: SecKeychainSearchCopyNext: The specified item could not be found in the keychain.
```

the-real-adammork:
@carybakker  keys re-encrypted with your public key, can you test that it worked with `scripts/decrypt_secrets` ?

thanks

carybakker:
I was able to decrypt the values successfully, but when running `scripts/reinitialize_keys`, I get the following error:
```
scripts/reinitialize_keys: line 15: generate_and_rename_keys: command not found
```
Guessing the script just needs to be committed?

lmk

image

the-real-adammork:
@carybakker updated function name, should resolve that issue

scripts/decrypt_secrets
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
```
carybakker, line 2: the use of `set -e` caused a problem when one of the functions made a call that returned a non-zero return code, causing the script to terminate prematurely (see the commit I made with the script fix).  A better way might be to not rely on `set -e` and to instead explicitly check calls when needed and exit on failure so the impacted function that was fixed doesn't have to toggle `set -e` off and on... unless there's another fix for that issue that I'm not aware of.

scripts/secrets_manager
```diff
@@ -0,0 +1,138 @@
...
+
+function project_directory() {
+  script_path=$(readlink -f "${BASH_SOURCE[0]}")
+  script_dir=$(dirname "$script_path")
+  project_dir=$(dirname "$script_dir")
+  echo "$project_dir"
+}
+
+function check_for_existing_keys() {
+  set +e
```
carybakker, line 47: One problem I ran into was that I had generate the keys, but reset my codebase... So the script found the keys and exited without adding my public key to the file to be committed...  The workaround was that I had to manually delete the keys from my keychain.  A more robust check would be to ALSO check that the found public key exists in the file on the filesystem, and if it's not there, add it.  That would fix the above issue.

scripts/decrypt_secrets
the-real-adammork: `set -e` makes sense here

carybakker (commented):
Just a couple of comments added.  Will let Alex or someone above my paygrade approve since it involves committing encrypted data to the repo (i.e. they should approve the data stored, encryption used, and stipulate any process around managing the data if necessary - like, how are we going to be sure something more sensitive isn't added in the future without approval, or is it considered secure enough for us to put whatever we want - that kind of thing should ideally be documented somewhere instead of just relying on code reviews in github where it can be easy for someone to miss in a PR review).

scripts/secrets_manager
the-real-adammork: good idea, will implement this suggestion

the-real-adammork: added

voloshyn (approved):
LGTM. As long as the committed data is limited to devnets & testnet config/AccountKeys.

PR #173, opened 2022-07-30
Async/Await API wrappers for MobileCoinClient

plarson:
### Motivation

Unfortunately this only works on iOS 13+.

Using `withCheckedThrowingContinuation` it is possible to add some quick async/await wrappers to the `MobileCoinClient` API.

This allows a more linear API style with less indentation and callbacks.

```
let balance = try await client.updateBalances()
let transaction = try await client.prepareTransaction(
    to: recipient,
    amount: Amount(100, in: .MOB),
    fee: McConstants.DEFAULT_MINIMUM_FEE
)
try await client.submitTransaction(transaction: transaction.transaction)
let status = try await client.status(of: transaction.transaction)
```

### In this PR

* Add `MobileCoinClient+Async.swift` with wrappers for current public API.
* Add async/await version of the tests.

### DIFF
--- a/Sources/MobileCoinClient+Async.swift
+++ b/Sources/MobileCoinClient+Async.swift
@@ -10,13 +10,8 @@ extension MobileCoinClient {
     @discardableResult
     public func updateBalances() async throws -> Balances {
         try await withCheckedThrowingContinuation { continuation in
-            updateBalances { result in
-                switch result {
-                case .success(let value):
-                    continuation.resume(returning: value)
-                case .failure(let error):
-                    continuation.resume(throwing: error)
-                }
+            updateBalances {
+                continuation.resume(with: $0)
             }
         }
     }
@@ -24,13 +19,8 @@ extension MobileCoinClient {
     @discardableResult
     public func blockVersion() async throws -> BlockVersion {
         try await withCheckedThrowingContinuation { continuation in
-            blockVersion { result in
-                switch result {
-                case .success(let value):
-                    continuation.resume(returning: value)
-                case .failure(let error):
-                    continuation.resume(throwing: error)
-                }
+            blockVersion {
+                continuation.resume(with: $0)
             }
         }
     }
@@ -45,13 +35,8 @@ extension MobileCoinClient {
             prepareTransaction(to: recipient,
                                memoType: memoType,
                                amount: amount,
-                               fee: fee) { result in
-                switch result {
-                case .success(let value):
-                    continuation.resume(returning: value)
-                case .failure(let error):
-                    continuation.resume(throwing: error)
-                }
+                               fee: fee) {
+                continuation.resume(with: $0)
             }
         }
     }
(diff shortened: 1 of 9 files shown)


the-real-adammork:
this is rad

the-real-adammork:
Screen Shot 2022-07-30 at 2 01 46 PM
Screen Shot 2022-07-30 at 2 02 14 PM

do you know why the async failures prints a raw integer for the error type instead of the description like the other calls ?

plarson:
> do you know why the async failures prints a raw integer for the error type instead of the description like the other calls ?

I wasn't sure where this description came from, since it's catching the exceptions instead of getting the error from a Result. I think the 1 might be XCTest.

the-real-adammork:
> > do you know why the async failures prints a raw integer for the error type instead of the description like the other calls ?
> 
> I wasn't sure where this description came from, since it's catching the exceptions instead of getting the error from a Result. I think the 1 might be XCTest.

I think the 1 maps to second case in that error enum, is it possible to get it to show the description ?

plarson:
> > > do you know why the async failures prints a raw integer for the error type instead of the description like the other calls ?
> > 
> > 
> > I wasn't sure where this description came from, since it's catching the exceptions instead of getting the error from a Result. I think the 1 might be XCTest.
> 
> I think the 1 maps to second case in that error enum, is it possible to get it to show the description ?

I think that's coming from somewhere in XCTest, this method fixes the messages:

`XCTFail((error as CustomStringConvertible).description)`

plarson:
> > > > do you know why the async failures prints a raw integer for the error type instead of the description like the other calls ?
> > > 
> > > 
> > > I wasn't sure where this description came from, since it's catching the exceptions instead of getting the error from a Result. I think the 1 might be XCTest.
> > 
> > 
> > I think the 1 maps to second case in that error enum, is it possible to get it to show the description ?
> 
> I think that's coming from somewhere in XCTest, this method fixes the messages:
> 
> `XCTFail((error as CustomStringConvertible).description)`

There is a comment about that message here https://forums.swift.org/t/pitch-xctassertnothrow-with-result-validation/19647/4

the-real-adammork (approved):
everything looks great !

Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
```diff
@@ -25,4 +25,22 @@ extension XCTestCase {
             sleep(transportProtocol == last ? 0 : interval)
         }
     }
+
+    @available(iOS 13.0, *)
+    func testSupportedProtocols(
+                description: String,
+                timeout: Double = 40.0,
```
carybakker: Couple of questions on this part - GRPC and HTTP currently have different timeouts...only thing I think we need for the tests is that the test timeout is slightly longer than the protocol timeout so the calls can legitimately fail prior to the test timeout for expected error conditions but I don't see where timeout or description are actually used here. If we could base the default on the greater of the two values plus a slight increment, that'd likely be best.

For the interval delay, do we actually need it?  Will add +10s  for every protocol related test, right?

Sources/MobileCoinClient+Async.swift
```diff
@@ -0,0 +1,77 @@
+//
```
carybakker, line 1: nice :)

carybakker (commented):
nice improvements :)

Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
the-real-adammork: 10s was existing configuration from the non-async version of this function. I think 10s is the right amount to make sure we don't run into race conditions. For local-dev, you can use the ExampleHTTP project and there will be no interval because theres only one protocol.

plarson: Updated with timeout logic

PR #175, opened 2022-07-31
Update Podfiles for Xcode 13.3+

plarson:
### Motivation

Lint and Performance Test builds would break with Xcode 13.4.1. I'm unsure how [the previous commit](https://github.com/mobilecoinofficial/MobileCoin-Swift/pull/173) passed. I did some investigation and this is a common issue, the fix is to increase the versions and disable Bitcode.

Bitcode isn't needed for Example apps that aren't submitted to the App Store, it's also been [deprecated](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes). The deployment version can be changed if needed.

### In this PR

* Update Example apps to iOS 12.2
* Remove minimum deployment versions in Example apps.
* Disable Bitcode
* Add `.DS_Store` to gitignore.


carybakker (approved):
LGTM

the-real-adammork (changes requested):
To update the minimum deployment target, we need to verify that Signal will not be affected.

plarson:
> To update the minimum deployment target, we need to verify that Signal will not be affected.

Yep makes sense, I set it to the same as Signal.

plarson:
@the-real-adammork this only changes the Example apps. The primary podspec is still the same iOS 10.

the-real-adammork:
> @the-real-adammork this only changes the Example apps. The primary podspec is still the same iOS 10.

ok ya then this is fine.

the-real-adammork (approved):

PR #177, opened 2022-08-02
Gemfile Cleanup - Remove empty `specs:`

the-real-adammork:
### Motivation

Remove empty "specs:" sections, that bundler is removing upon project build. 

### In this PR
* Gemfile.lock chnages

### Future Work
* Update CICD?


the-real-adammork:
had wrong version of bundler installed upgrading my local solved this issue.

PR #179, opened 2022-08-10
Support for Transaction Idempotence

carybakker:
Soundtrack of this PR: [Oops!...I Did It Again](https://www.youtube.com/watch?v=CduA0TULnow)

### Motivation

To support re-trying transaction submission without violating idempotency, Moby (and other clients) need the ability to re-create identical transaction data via the `prepareTransaction` family of methods on MobileCoinClient.

### In this PR
* Added ability to provide a deterministic RNG to prepareTransaction methods, so the client can re-create the rng state prior to the call, if needed, in order to deterministically re-create identical transaction data.

Pivotal Tracker: [[iOS] Moby Transaction Idempotence](https://www.pivotaltracker.com/story/show/182820695)


Sources/MobileCoinClient.swift
```diff
@@ -217,11 +218,73 @@ public final class MobileCoinClient {
         }
     }
 
+    public func prepareTransaction(
```
the-real-adammork, line 210: I lean towards not adding these extra functions to the public API, we could modify the existing API and provide a default value in the function signature. Anyone currently calling the api won't have to change anything.

Sources/Utils/Rng/MobileCoinSeedableRng.swift
```diff
@@ -0,0 +1,50 @@
...
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+// swiftlint:disable unused_setter_value
+import Foundation
+
+public class MobileCoinSeedableRng: MobileCoinRng {
+    let _seed: Data32
+
+    var seed: Data32 {
```
the-real-adammork: don't think this dual variable pattern is necessary if the seed is declared as let, means it can only be written to once.

Sources/Transaction/Inputs/MixinSelectionStrategy.swift
```diff
@@ -26,3 +26,56 @@ extension MixinSelectionStrategy {
...
+            ringSize: McConstants.RING_SIZE)
+    }
+
+    func selectMixinIndices(
+        forRealTxOutIndices realTxOutIndices: [UInt64],
+        selectionRange: PartialRangeUpTo<UInt64>?,
+        excludedTxOutIndices: [UInt64] = []
+    ) -> [Set<UInt64>] {
+        selectMixinIndices(
+            rng: MobileCoinDefaultRng(),
```
the-real-adammork, line 61: feel like the `IdempotentMixinSelectionStragegy` should use the chacha20 RNG by default

Sources/Transaction/Inputs/DefaultMixinSelectionStrategy.swift
```diff
@@ -4,16 +4,21 @@
 
 import Foundation
 
-final class DefaultMixinSelectionStrategy: MixinSelectionStrategy {
+final class DefaultMixinSelectionStrategy: IdempotentMixinSelectionStragegy {
```
the-real-adammork: for backwards compatibility lets keep this file as-it-was, and make a new selection strategy implementation that inherits from `IdempotentMixinSelectionStragegy`. 

I could be persuaded to migrate the default implementation to chacha20/seedableRNG, but my initial thought is to keep the existing unchanged, and add.

the-real-adammork: ok actually I think I have better idea. will test out in a separate branch.

the-real-adammork: actually, I think your solution is correct, users get existing behavior from the extension

Sources/Transaction/Inputs/MixinSelectionStrategy.swift
the-real-adammork: nevermind, I understand how this works now

Sources/Network/HTTPS/HttpProtocolConnectionFactory.swift
```diff
@@ -15,7 +15,7 @@ class HttpProtocolConnectionFactory: ProtocolConnectionFactory {
         config: AttestedConnectionConfig<ConsensusUrl>,
         targetQueue: DispatchQueue?,
         rng: (@convention(c) (UnsafeMutableRawPointer?) -> UInt64)?,
-        rngContext: Any?
+        rngContext: Any? = nil
```
the-real-adammork: is this a temporary change ? I feel like providing a default value of nil can lead to programmer error, if we need still need this parameter we should remove the default of nil and make sure all call-sites pass whats correct, or if its not needed we remove the parameter from the function

Sources/Transaction/Inputs/DefaultMixinSelectionStrategy.swift
```diff
@@ -4,16 +4,21 @@
...
     let offsetParam: UInt64 = 88
-    var rng: RandomNumberGenerator = SystemRandomNumberGenerator()
+    var rng: MobileCoinRng
+
+    init(rng: MobileCoinRng = MobileCoinDefaultRng()) {
```
the-real-adammork: do we need this on the init if the function calls require it to be passed in ?

Sources/Utils/Rng/MobileCoinRng.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+// swiftlint:disable unavailable_function
+public class MobileCoinRng {
```
the-real-adammork: what about making this conform to the Swift standard protocol `RandomNumberGenerator` and renaming the nextUInt64() to next(), the return types are the same.

Sources/Utils/Rng/MobileCoinSeedableRng.swift
```diff
@@ -0,0 +1,50 @@
...
+
+        guard let seedData32 = Data32(Data(randomBytes)) else {
+            // will not happen
+            logger.fatalError("Failed to generate Data32 for rng seed.")
+        }
+
+        self.init(seed: seedData32)
+    }
+
+    var wordPos: Data16 {
```
the-real-adammork: if this isnt available on the default implementation, lets remove it. Internally its not needed so it doesn't need to be part of the base-class. SDK consumers that need this from the ChaCha20 implementation can still get it because the reference they hold onto will be typed.

Sources/Utils/Rng/MobileCoinSeedableRng.swift
```diff
@@ -0,0 +1,50 @@
...
+    var seed: Data32 {
+        _seed
+    }
+
+    init(seed: Data32) {
+        self._seed = seed
+    }
+
+    override convenience init() {
+        let bytesCount = 32
```
the-real-adammork: first part of this code looks the same as `func securityRNG(...)` maybe just re-use that to avoid duplication ?

Sources/Utils/Rng/MobileCoinChaCha20Rng.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+import LibMobileCoin
+
+final class MobileCoinChaCha20Rng: MobileCoinSeedableRng {
```
the-real-adammork: looking at these FFI calls - https://github.com/mobilecoinfoundation/mobilecoin/blob/f3eff68f1c8bf53e8635c63b10bc79197e4b1848/libmobilecoin/src/chacha20_rng.rs - the `unwrap()`s & `expect()`s will cause rust panics. The pattern for dealing with errors can be seen in other rust files, and their corresponding swift wrappers. Change the FFI calls, and these call-sites to catch and "re-throw" if needed.

Sources/Utils/Rng/MobileCoinChaCha20Rng.swift
```diff
@@ -0,0 +1,48 @@
...
+                mc_chacha20_rng_create_with_bytes(bytesBufferPtr)
+            }
+        }
+        ptr = withMcInfallible {
+            cc20ptr
+        }
+        super.init(seed: seed)
+    }
+
+    override var wordPos: Data16 {
```
the-real-adammork: I feel like this is too much "work"/complexity for a computed variable, should be a function, and that also allows for a `throw` modifier (if needed, see above comment).

Tests/Integration/Network/Protocol/TransactionIdempotenceTests.swift
```diff
@@ -0,0 +1,113 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+@testable import MobileCoin
+import XCTest
+
+class TransactionIdempotenceTests: XCTestCase {
```
the-real-adammork: Looks like this ended up in the wrong folder in the repo, but also does this need to be an integration test ? seems like its better suited as a unit test case.

the-real-adammork:
Screen Shot 2022-08-11 at 10 47 03 PM

I know this is still WIP, but the unit tests fail for me with this error, might have something todo with how the default implementation has changed.

Sources/LibMobileCoin/McRngCallback.swift
```diff
@@ -5,15 +5,34 @@
...
 func withMcRngCallback<T>(
     rng: (@convention(c) (UnsafeMutableRawPointer?) -> UInt64)?,
     rngContext: Any?,
     _ body: (UnsafeMutablePointer<McRngCallback>?) throws -> T
 ) rethrows -> T {
     if let rng = rng {
-        var rngContext = rngContext
-        return try withUnsafeMutablePointer(to: &rngContext) { rngContextPtr in
-            var rngCallback = McRngCallback(rng: rng, context: rngContextPtr)
+        if let rngContext = rngContext {
```
the-real-adammork: edited: don't think this change is needed, current code works with out it

Sources/MobileCoinClient.swift
```diff
@@ -204,6 +204,7 @@ public final class MobileCoinClient {
             metaFetcher: metaFetcher,
             txOutSelectionStrategy: txOutSelectionStrategy,
             mixinSelectionStrategy: mixinSelectionStrategy,
+            rng: MobileCoinDefaultRng(),
```
the-real-adammork: thinking maybe its better to have one instance of the default RNG that we keep attached to the client.

Tests/Common/Utils/TestRng.swift
```diff
@@ -38,4 +38,17 @@ func testRngCallback(context: UnsafeMutableRawPointer!) -> UInt64 {
     context.assumingMemoryBound(to: Xoshiro.self).pointee.next()
```
the-real-adammork:
```suggestion
    context.assumingMemoryBound(to: MobileCoinXoshiroRng.self).pointee.nextUInt64()
```

the-real-adammork: this fixes the error I posted in that screenshot

Sources/LibMobileCoin/McRngCallback.swift
```diff
@@ -5,15 +5,34 @@
 import Foundation
 import LibMobileCoin
 
+func withMcRngObjCallback<T>(
```
the-real-adammork, line 8: some thoughts on this change:

if we're moving all RNG behavior to MobileCoinRng sub-classes, then the "rng function" that accepts the context will be the same because it relies on common behavior of the base class. Given that, I don't think it needs to be parameterized here, rather it can just be set explicitly.

Tests/Unit/Transaction/TransactionTests.swift
```diff
@@ -80,7 +83,8 @@ class TransactionTests: XCTestCase {
             fee: fixture.fee,
             tombstoneBlockIndex: fixture.tombstoneBlockIndex,
             fogResolver: fixture.fogResolver,
-            blockVersion: .minRTHEnabled).get()
+            blockVersion: .minRTHEnabled,
+            rng: TestRng()).get()
```
the-real-adammork, line 87: nit: this change isn't necessary because it should be able to use a default value right ?

Tests/Integration/Network/Protocol/TransactionIdempotenceTests.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    func testTransactionIdempotence() throws {
+        let seed = Data32(repeating: 5)
```
the-real-adammork: some duplication in both of these test-cases, could be abstracted, or maybe try using the new async calls that phil added. those will help with the nesting hell.

Sources/Utils/SecurityRNG.swift
```diff
@@ -25,3 +25,15 @@ func securityRNG(context: UnsafeMutableRawPointer?) -> UInt64 {
 
     return value
 }
+
+func mobileCoinRNG(context: UnsafeMutableRawPointer?) -> UInt64 {
+    // get MobileCoinChaCha20Rng from context
```
the-real-adammork:
```suggestion
    // get MobileCoinRng sub-class from context    
```

the-real-adammork:
high-level, looks good and seems to work as expected. Couple thoughts:

- Should we convert our default RNG for transactions to be ChaCha20 w/ a random seed ?
- I see RNG used in some other non-transaction building places, should we convert those too ? (don't see this change on android tho)

API-wise, what we have works and it matches android but I feel like asking consumers to create their own ChaCha20Rng references, manage, and then continually pass it in to the public api is extra work. I suggest a future enhancement where the public RNG parameter is an enum like:

```swift
enum RngProvider {
case standard
case chacha20(seed: UInt64, position: Data16)
case custom(provider: MobileCoinRng)
}
```

then we could modify both `PendingTransaction` & `PendingSinglePayloadTransaction` with an "rngContext" variable containing the "position". This assumes the SDK consumer is keeping track of (or knows the source of truth) for their seed. This is enough for idempotence and deletes responsibility of managing RNG instances.

We'd also have to add a function to the `MobileCoinRng` base-class that returns an "rngContext", but that would be simple. That way consumers can implement their own seedable/contextual RNGs. 

correction: we could create a protocol `ConextualRandomNumberGenerator` w/ `func getContext() -> Any?`. For RNGs that don't conform we'd set a value of nil on the PendingTransaction structs for "rngContext".

Sources/Utils/Rng/MobileCoinSeedableRng.swift
carybakker: Ya, I think I originally had it as var with set/get before I realized that there was no setter on the underlying rust impl.
Will update

Sources/Transaction/Inputs/DefaultMixinSelectionStrategy.swift
carybakker: Left over from earlier code pass where I had it as an attribute... nice catch.

carybakker: code updated

Sources/Utils/Rng/MobileCoinSeedableRng.swift
carybakker: Removed MobileCoinSeedableRng abstraction as it doesn't really seem necessary to have it for now.. but I can put back with just the seed storage if you think it'd be useful to still have the seedable abstraction (seems like we only really need the base MobileCoinRng)

Sources/Utils/Rng/MobileCoinSeedableRng.swift
carybakker: The `func securityRNG(...)` func only returns a UInt64 / 8 bytes, and has a `context` param that's really only relevant for it's use as a callback...   But maybe we can refactor the implementation of securityRNG AND this code to both use some convenience function(s) to generate data of a specified size and have them both call that function... that way the RNG generation is common between both.  Will give it a shot.

Sources/Utils/Rng/MobileCoinSeedableRng.swift
the-real-adammork: I think its ok to remove.

I made a suggestion at the end about encapsulating the RNG implementations, the "contextual/seedable" behavior of an RNG could be defined in a protocol, and then RNGs can choose to implement. For this example and chacha20, the context function would return a Data16 representing the word position, this would get set on the PendingTransaction structs returned to consumers, value would be nil for RNGs that don't implement this behavior.

Tests/Integration/Network/Protocol/TransactionIdempotenceTests.swift
carybakker: These are calls that go through to the mobilecoin/libmobilecoin APIs, so it's not really a "unit" test per-se.
Depends on how we categorize tests...

But yes, definitely didn't mean for them to be under Network/Protocol

Sources/LibMobileCoin/McRngCallback.swift
carybakker: Right... I originally started out sharing the same `func withMcRngCallback<T>(...)` func when it seemed like 'context' wasn't being used anywhere (I didn't notice the test code).  I should have removed this when I switched to a different func to explicitly handle passing in MobileCoinRng.

Code updated.

Sources/MobileCoinClient.swift
carybakker: Ah, yeah, I think you're right.  When I switched to passing param to the `prepare` calls I pulled that out (I originally had it set up to be attached to the client when just getting things working).

Will update to default to using client's RNG (which will be MobileCoinDefaultRng()) when no rng is passed as a param

Tests/Common/Utils/TestRng.swift
carybakker: Thanks -

Sources/Utils/Rng/MobileCoinSeedableRng.swift
carybakker: This is moved into the `MobileCoinChaCha20Rng` class for now, after removing the `MobileCoinSeedableRng`, and the code has been updated to remove the dual varaible.

Sources/MobileCoinClient.swift
carybakker: Sounds good - I was kind of thinking the same...

carybakker: consolidated rng & non-rng APIs via optional param defaulting to nil, where implementation will use the internal MobileCoinDefaultRng().

Sources/Network/HTTPS/HttpProtocolConnectionFactory.swift
carybakker: Optional values default to nil regardless...  However, I believe I had to add in the setting to nil explicitly for some reason... will re-check this to see, but I ran into a similar thing just now when consolidating the MobileCoinClient.prepare* calls... where when the optional is not explicitly set, the compiler was having trouble identifying things, and seemed be trying to match up the params with the deprecated calls (i.e. it was complaining that Amount was not the expected type of UInt64)

Sources/Utils/Rng/MobileCoinRng.swift
carybakker: Sounds good - will do.

Sources/Network/HTTPS/HttpProtocolConnectionFactory.swift
the-real-adammork: optional values do not default to nil, but nil is an acceptable value for the parameter. My comment is basically "we should force call-sites to be explicit about what they want the value to be, if they want nil they can pass nil."

Tests/Integration/Network/Protocol/TransactionIdempotenceTests.swift
the-real-adammork: Tests that need network/environment access are Integration tests, if they don't need the network, the default is unit test.

Sources/MobileCoinClient.swift
carybakker: updated.

carybakker:
> * Should we convert our default RNG for transactions to be ChaCha20 w/ a random seed ?

Good question... I'm not sure, that's more of a crypto side question.. I'll have to double-check the Java side, b/c I'm pretty sure Bernie told me or I read in a Slack convo between Chris/Bernie that they were changing to use a new ChaCha20Rng seeded with 32 bytes from the non-deterministic / original Rng...

carybakker:
> then we could modify both PendingTransaction & PendingSinglePayloadTransaction with an "rngContext" variable containing the "position". 

I like this idea... I was thinking of something similar, basically wrapping things up with some kind of opaque type that abstracts things...

What I'm thinking now that might be really nice for clients.... is for the opaque 'context' that the user needs to re-create a transaction to be provided directly by the transaction itself.

If we do that, AND update the logic to always use ChaCha20Rng (seeded with a new, random 32 bytes), we can cache all that stuff and the user can just get that 'context' from the transaction to re-create.... 

This way the client doesn't need to do anything by default - APIs are identical.  But when re-trying, they get the opaque context from the failed transaction and pass that in to the prepare call via the optional opaque context param, and voila - new tx ... wdyt?

Sources/Utils/Rng/MobileCoinSeedableRng.swift
carybakker: > I made a suggestion at the end about encapsulating the RNG implementations

Added some thoughts in a comment after that too that kind of evolve that idea a bit too... marking this resolved, and we can pick up the convo down there.

Sources/Utils/Rng/MobileCoinRng.swift
carybakker: code updated.

the-real-adammork:
the Context only works if they're also providing the seed, and I dont think that will be the default. 

My suggestion puts them in the struct you get back from prepare, so that info can be cached same as if it was in the transaction. Lets revisit your idea If/when we implement auto-retry/rebuild at the SDK level.

Sources/Network/HTTPS/HttpProtocolConnectionFactory.swift
carybakker: Ah, yes, you're correct, I think I was left over from my initial pass at wiring things up... not sure, but it's definitely not necessary to default to nil.

Sources/LibMobileCoin/McRngCallback.swift
carybakker: Nice catch - removed param and setting explicitly for MobileCoinRng usage.

Sources/Utils/Rng/MobileCoinSeedableRng.swift
carybakker: DRY'ed this out via an extension on Data to generate a specified number of random bytes of data, and plugged that in for both the default func for the UInt64 and for the generation of the 32 byte seed for ChaCha20

Tests/Integration/Network/Protocol/TransactionIdempotenceTests.swift
carybakker: Understood.  Moved from Network/Protocol to Transaction/ for now (still under Integration).  Would have to redo a lot of wiring for them to work b/c they're leveraging the existing fixtures for integration tests where the accounts have data / transactions for creating the input and change transaction...  Might not take that long, but I can add a follow up task to move this from integration to unit testing after the code is merged.

Sources/Utils/Rng/MobileCoinChaCha20Rng.swift
carybakker: Moved to function - will update with throw if needed when addressing prior comment

the-real-adammork:
Mikey left some feedback on the rust PR, based on that maybe it makes sense to parameterize the rust calls and not hold onto the chacha20 reference that would solve for the lock issue he mentioned and make the calls more transactional. Would also solve for the hesitation I had to let consumers hold onto the RNG reference themselves.

Sources/LibMobileCoin/McError.swift
```diff
@@ -41,6 +41,27 @@ func withMcInfallible(_ body: () -> Bool) {
     }
 }
 
+func withMcError<T>(_ body: (inout UnsafeMutablePointer<McError>?) -> T)
+    -> Result<(T), LibMobileCoinError>
+{
+    var error: UnsafeMutablePointer<McError>?
+    let value = body(&error)
+
+    if let mcError = error {
```
the-real-adammork, line 50: can we make this a guard ?

Sources/LibMobileCoin/McError.swift
```diff
@@ -41,6 +41,27 @@ func withMcInfallible(_ body: () -> Bool) {
...
+        let err: LibMobileCoinError
+        do {
```
the-real-adammork, line 52: this could be collapsed into a guard with a `try?`

Sources/MobileCoinClient.swift
```diff
@@ -193,17 +194,21 @@ public final class MobileCoinClient {
         memoType: MemoType = .recoverable,
         amount: Amount,
         fee: UInt64,
+        rng: MobileCoinRng? = nil,
```
the-real-adammork: why not make this parameter non-optional with a default value of `defaultRng` ? Removes the nil coalescing operator stuff below, and gives the same result.

Sources/Network/Service/DefaultServiceProvider.swift
```diff
@@ -22,7 +22,8 @@ final class DefaultServiceProvider: ServiceProvider {
         networkConfig: NetworkConfig,
         targetQueue: DispatchQueue?,
         grpcConnectionFactory: GrpcProtocolConnectionFactory,
-        httpConnectionFactory: HttpProtocolConnectionFactory
+        httpConnectionFactory: HttpProtocolConnectionFactory,
+        rng: MobileCoinRng? = nil
```
the-real-adammork: why do we need to make changes here ? I see that you're adding this parameter but its not used anywhere.

Sources/LibMobileCoin/McError.swift
```diff
@@ -41,6 +41,27 @@ func withMcInfallible(_ body: () -> Bool) {
     }
 }
 
+func withMcError<T>(_ body: (inout UnsafeMutablePointer<McError>?) -> T)
+    -> Result<(T), LibMobileCoinError>
```
plarson: Maybe this should just be `Result<T, LibMobileCoinError>` instead of a Tuple with one element?

Sources/Transaction/Inputs/MixinSelectionStrategy.swift
```diff
@@ -26,3 +26,56 @@ extension MixinSelectionStrategy {
             ringSize: McConstants.RING_SIZE)
     }
 }
+
+protocol IdempotentMixinSelectionStragegy: MixinSelectionStrategy {
```
plarson: Typo `IdempotentMixinSelectionStrategy`

the-real-adammork: nitpick: If we're gonna call this protocol "Idempotent" it should only use  "Contextual" RNGs. I made a suggestion to make a protocol "ContextualRNG" in another comment, which we could do, and then require here, OR the quick fix is to require MobileCoinChaCha20Rng as the type here because know it supports idempotence, whereas just a MobileCoinRng does not.

Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
```diff
@@ -14,7 +14,7 @@ extension XCTestCase {
                 interval: UInt32 = 10,
                 _ testCase: (TransportProtocol, XCTestExpectation) throws -> Void
     ) rethrows {
-        let supportedProtocols = TransportProtocol.supportedProtocols
+        let supportedProtocols = [TransportProtocol.http]// TransportProtocol.supportedProtocols
```
plarson: Is this a permanent change?

Sources/Utils/Rng/MobileCoinChaCha20Rng.swift
```diff
@@ -0,0 +1,137 @@
...
+                    // non-documented errors.
+                    logger.fatalError("Unhandled LibMobileCoin error: \(redacting: error)")
+                }
+            }
+        }
+
+        self.seed = seed32.data
+    }
+
+    convenience init(seed: Data = Data.secRngGenBytes(32)) {
```
the-real-adammork:
```suggestion
    convenience init(seed: Data = .secRngGenBytes(32)) {
```

Sources/Utils/Data/Data+SecRng.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+
+import Foundation
+
+extension Data {
+    public static func secRngGenBytes(_ count: Int) -> Data {
+        Data(withFixedLengthMcMutableBufferInfallible: count) {
+            SecRandomCopyBytes(kSecRandomDefault, count, $0.pointee.buffer) == errSecSuccess
```
the-real-adammork, line 10: nit: the error printed here would say `"Infallible LibMobileCoin function failed."` which is a little misleading because its not coming from a LibMobileCoin FFI call. 

Maybe make a new one with a different error message, or change the existing to be more generic like`"Infallible Unsafe Mutable Data Allocation failed."`

Tests/Unit/Rng/SeedableRngUnitTests.swift
```diff
@@ -0,0 +1,34 @@
...
+
+        XCTAssertNotNil(rng1)
+        XCTAssertNotNil(rng2)
+        XCTAssertNotNil(rng3)
+
+        let val1 = rng1.next()
+        let val2 = rng2.next()
+        let val3 = rng3.next()
+
+        print("********** SEEDABLE RNG UNIT TESTS")
```
the-real-adammork: Seems a little debug-ey, if this is meant to stay, lets make it more readable.

the-real-adammork:
RE: My suggestion to enumerate RNG types and your idea for better retry, lets park these all as "future improvements" I think what you have is sufficient for now, we can revisit these type of things in the future after merge.

Sources/Transaction/Inputs/MixinSelectionStrategy.swift
the-real-adammork: Or change the protocol name to something like `CustomRNGMixinSelectionStrategy` to reduce confusion

Sources/LibMobileCoin/McError.swift
the-real-adammork:
```suggestion
    -> Result<T, LibMobileCoinError>
```

Sources/MobileCoinClient.swift
the-real-adammork:
```suggestion
        rng: MobileCoinRng = defaultRng,
```

Sources/Transaction/TransactionPreparer.swift
```diff
@@ -148,10 +153,20 @@ struct TransactionPreparer {
         merkleRootBlock: UInt64? = nil,
         completion: @escaping (Result<[PreparedTxInput], ConnectionError>) -> Void
     ) {
-        let inputsMixinIndices = mixinSelectionStrategy.selectMixinIndices(
-            forRealTxOutIndices: inputs.map { $0.globalIndex },
-            selectionRange: ledgerTxOutCount.map { ..<$0 }
-        ).map { Array($0) }
+        var inputsMixinIndices: [[UInt64]]
+
+        if let idempotentStrategy = mixinSelectionStrategy as? IdempotentMixinSelectionStragegy {
```
the-real-adammork: naming change possibly

the-real-adammork:
Future Improvements:
- Move integration to unit test with mocks
- Contextual RNG Protocols and enum
- Improved coverage of error cases on the new FFIs

Sources/MobileCoinClient.swift
carybakker: @the-real-adammork  So we can do this, but we have to make the defaultRng attribute of MobileCoinClient public static for it to be used as a default param value.  Might be some other option - will explore further.  For now will update to make it work and let you take a look.

Sources/Network/Service/DefaultServiceProvider.swift
carybakker: Ya, these are part of a bunch of vestigial changes from when I had tested the ChaCha20Rng as a general replacement for the default RNG.  Cleaning them out.

Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
carybakker: haha - no! :)

Tests/Unit/Rng/SeedableRngUnitTests.swift
carybakker: ah, yes.. this was definitely debuggy... cleaning up

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "Vendor/libmobilecoin-ios-artifacts"]
 	path = Vendor/libmobilecoin-ios-artifacts
-	url = https://github.com/mobilecoinofficial/libmobilecoin-ios-artifacts.git
```
the-real-adammork: this should be reverted

carybakker: will update when libmobilecoin-ios-artifacts published

carybakker: ah, looks like it's not working... will revert

Sources/Transaction/TransactionPreparer.swift
carybakker: updating...

Sources/Network/Connection/Connections/ConsensusConnection.swift
```diff
@@ -15,23 +15,17 @@ final class ConsensusConnection: Connection<
     private let grpcFactory: GrpcProtocolConnectionFactory
     private let config: NetworkConfig
     private let targetQueue: DispatchQueue?
-    private let rng: (@convention(c) (UnsafeMutableRawPointer?) -> UInt64)?
-    private let rngContext: Any?
```
the-real-adammork, line 19: Why do we need to remove these ?

the-real-adammork: Look at all XXXConnection classes

Sources/Network/Connection/ProtocolConnectionFactory.swift
```diff
@@ -17,9 +17,7 @@ protocol ProtocolConnectionFactory {
 
     func makeConsensusService(
         config: AttestedConnectionConfig<ConsensusUrl>,
-        targetQueue: DispatchQueue?,
```
the-real-adammork, line 20: here too

Sources/Transaction/TransactionBuilder.swift
```diff
@@ -52,7 +52,8 @@ final class TransactionBuilder {
                             tombstoneBlockIndex,
                             fogResolverPtr,
                             memoBuilderPtr,
-                            blockVersion)
+                            blockVersion,
+                            nil)
```
the-real-adammork, line 56: whats this change for ?

the-real-adammork: the header file indicates that this parameter is `McError* MC_NULLABLE * MC_NULLABLE out_error`

so should be handled

the-real-adammork: https://github.com/mobilecoinofficial/libmobilecoin-ios-artifacts/blob/release/v2/Artifacts/include/transaction.h#L193

the-real-adammork: we need this change https://github.com/mobilecoinofficial/MobileCoin-Swift/pull/178

Tests/Integration/Transaction/TransactionIdempotenceTests.swift
```diff
@@ -0,0 +1,108 @@
+//
+//  Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+//
+// swiftlint:disable empty_xctest_method
+
+@testable import MobileCoin
+import XCTest
+
+#if swift(>=5.5)
```
the-real-adammork, line 9: fix needed before approval

the-real-adammork:
push this to `mobilecoinofficial` and re-open

MobileCoin.podspec
```diff
@@ -58,7 +58,7 @@ Pod::Spec.new do |s|
       "Sources/Network/*.{h,m,swift}",
```
the-real-adammork, line 58: update line:6 `s.version = "1.2.4"` -> `s.version = "2.0.0"`

and update the changelog.md

carybakker:
Replaced by PR#188 from task/idempotence-release branch

MobileCoin.podspec
carybakker: done

carybakker: (done in PR188)

Tests/Integration/Transaction/TransactionIdempotenceTests.swift
carybakker: Fixed in P R #188, reverted to old nested closure format w/out async.

Sources/Network/Connection/ProtocolConnectionFactory.swift
carybakker: in PR #188, I restored all the rng code that was in place prior to starting idempotence.

Sources/Network/Connection/Connections/ConsensusConnection.swift
carybakker: in PR #188, I restored all the rng code that was in place prior to starting idempotence.

PR #181, opened 2022-08-16
Add requireInBalance flag to TransactionStatusChecker

plarson:
Soundtrack of this PR: [Loading...](https://music.apple.com/us/album/loading-feat-hm-surf/1468495799?i=1468495800)

### Motivation

When sending transactions, sometimes a caller only needs to know that the txOut is found, and the keyImage is spent in ledger without checking the blocks included in the balance.

This balance check can make sending transactions appear slower since the balance needs to be updated before the transaction status can be checked.

### In this PR
* Adds `requireInBalance` flag to control whether `TransactionStatusChecker` feigns ignorance.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 defaults:
-  default-xcode-version: &default-xcode-version "13.2.1"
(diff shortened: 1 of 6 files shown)


the-real-adammork (approved):
looks good

PR #183, opened 2022-08-22
Add TransactionStatusTxOutChecker for simpler and quicker transaction status

plarson:
### Motivation

It's not always necessary to check the keyImage when fetching transaction status because the `FogLedger_TxOutResult` has the same information. For backwards compatibility this was added as a new class `TransactionStatusTxOutChecker`.

### In this PR
* Add `TransactionStatusTxOutChecker` which just relies on UntrustedTxOut service.

### DIFF
--- a/Sources/LibMobileCoin/ProtoExtensions.swift
+++ b/Sources/LibMobileCoin/ProtoExtensions.swift
@@ -186,3 +186,16 @@ extension FogLedger_BlockData {
         BlockMetadata(index: index, timestampStatus: timestampStatus)
     }
 }
+
+extension FogLedger_TxOutResult {
+    var timestampDate: Date? {
+        get { timestamp != UInt64.max ? Date(timeIntervalSince1970: TimeInterval(timestamp)) : nil }
+        set {
+            if let newValue = newValue {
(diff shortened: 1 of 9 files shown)


Sources/LibMobileCoin/ProtoExtensions.swift
```diff
@@ -186,3 +186,16 @@ extension FogLedger_BlockData {
         BlockMetadata(index: index, timestampStatus: timestampStatus)
     }
 }
+
+extension FogLedger_TxOutResult {
+    var timestampDate: Date? {
```
the-real-adammork, line 191: Whats the purpose of this extension ?

plarson: This is the same extension that exists is for [FogView_TxOutRecord](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/7c5dd2f5502c907ec82f92f414c4f38eb39b84a1/Sources/LibMobileCoin/ProtoExtensions.swift#L91) because the timestamp value is `UInt64.max` when it is not valid and `BlockMetadata` expects a `Date` object

the-real-adammork (approved):

PR #186, opened 2022-08-25
Stubs for latest libmobilecoin

plarson:
### Motivation

Moving libmobilecoin to another repo: https://github.com/mobilecoinofficial/libmobilecoin-ios-artifacts/pull/33

### In this PR
* Adds stubs for new symbols in libmobilecoin


Sources/Encodings/Base58Coder.swift
```diff
@@ -52,6 +52,8 @@ public enum Base58Coder {
                 return nil
             }
             return .transferPayload(transferPayload)
+        case.txOutGiftCode:
```
the-real-adammork:
```suggestion
        case .txOutGiftCode:
```

the-real-adammork (approved):

PR #190, opened 2022-08-30
Change @available rule to be compatible with Signals build process

the-real-adammork:
Soundtrack of this PR: [Yello - Oh Yeah](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwjfoab82u_5AhU9ADQIHRY-C_YQyCl6BAgbEAM&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3D6jJkdRaa04g&usg=AOvVaw0NvyoNJhkHOJNCcENbvLR4)

### Motivation

the `@available` tags need a higher version to be compatible with Signal's build process

### In this PR
* Change version and type of `@available` definitions

### DIFF
--- a/Sources/MobileCoinClient+Async.swift
+++ b/Sources/MobileCoinClient+Async.swift
@@ -6,7 +6,7 @@ import Foundation
 
 #if swift(>=5.5)
 
-@available(iOSApplicationExtension 15.0.0, *)
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 extension MobileCoinClient {
 
     @discardableResult
--- a/Sources/MobileCoinClient+Async.swift
+++ b/Sources/MobileCoinClient+Async.swift
@@ -6,7 +6,7 @@ import Foundation
 
 #if swift(>=5.5)
 
-@available(iOS 13.0, *)
+@available(iOSApplicationExtension 15.0.0, *)
 extension MobileCoinClient {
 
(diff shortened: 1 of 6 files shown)


Sources/MobileCoinClient+Async.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 
 #if swift(>=5.5)
 
-@available(iOS 13.0, *)
+@available(iOSApplicationExtension 15.0.0, *)
```
plarson: I think this should still be `iOS`

plarson: Since you're changing these, we might want to do the full specifier like GRPC does:

`@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)`

https://github.com/grpc/grpc-swift/blob/7b6857655d50b4c83cd2191b6135b432106e147c/Sources/GRPC/AsyncAwaitSupport/GRPCClient%2BAsyncAwaitSupport.swift#L20

the-real-adammork: ok ya I think you're correct, this change came from accepting the Xcode suggestion. Ill update it to iOS and add the others

stevenspiel (approved):
LGTM

plarson (approved):

PR #194, opened 2022-09-20
Simplify Idempotence Capability

the-real-adammork:
Soundtrack of this PR: [Idris Muhammed - Could Heaven Ever Be Like This](https://www.youtube.com/watch?v=0skigTyA4NE)

### Motivation

The requirements for Transaction Idempotence have changed, and now only require a Seed to be passed down into the Transaction Builder.

### In this PR
* Add RngSeed type so consumers can pass in a seed to the Transaction Builder
* Remove RNG seedability for txo selection
* Update unit tests
* Add new comparison utils for testing if TxOuts are idempotent (same txOutPublicKey)


Tests/Integration/Network/NetworkConfigFixtures.swift
```diff
@@ -10,7 +10,7 @@ import XCTest
 
 enum NetworkConfigFixtures {
     static let alphaDev = DynamicNetworkConfig.AlphaDevelopment.make()
-    static let network: NetworkPreset = .dynamic(alphaDev)
+    static let network: NetworkPreset = .testNet
```
plarson, line 13: Does this affect CI?

plarson (approved):

stevenspiel (approved):

Tests/Integration/Network/NetworkConfigFixtures.swift
the-real-adammork: no effect on CI, and should be testNet as thats the most stable test network

Sources/MobileCoinClient+Deprecated.swift
```diff
@@ -209,6 +209,92 @@ extension MobileCoinClient {
...
+            rngSeed: RngSeed,
+            completion: @escaping (
+                Result<PendingSinglePayloadTransaction, TransactionPreparationError>
+            ) -> Void
+        ) {
+        ```
+
+        this function creates a 32-byte seed by combining the data from 4 calls to rng.next()
+        """)
+    public func prepareTransaction(
```
alanpaulin, line 231: Is this for backwards compatibility so that Moby can integrate the updated SDK without simultaneously updating to use the new API? 👏

Sources/Transaction/Outputs/PendingTransaction.swift
```diff
@@ -35,3 +35,33 @@ public struct PendingSinglePayloadTransaction {
 
 extension PendingTransaction: Equatable, Hashable {}
 extension PendingSinglePayloadTransaction: Equatable {}
+
+extension PendingSinglePayloadTransaction {
+    func isIdempotent(_ lhs: PendingSinglePayloadTransaction) -> Bool {
+        Self.areIdempotent(self, lhs)
+    }
+
+    static func areIdempotent(
```
alanpaulin, line 44: Nice

Sources/Transaction/Outputs/PendingTransaction.swift
```diff
@@ -35,3 +35,33 @@ public struct PendingSinglePayloadTransaction {
 
 extension PendingTransaction: Equatable, Hashable {}
 extension PendingSinglePayloadTransaction: Equatable {}
+
+extension PendingSinglePayloadTransaction {
+    func isIdempotent(_ lhs: PendingSinglePayloadTransaction) -> Bool {
```
carybakker: given the meaning of idempotent, it's not 100% clear, by name alone, if this should compare for equality or inequality...   

Can we rename these to `isIdempotentlyEqual` or something like that? 

Same for the `are` functions...

alanpaulin (approved):

carybakker (approved):
LGTM!

Sources/Transaction/Outputs/PendingTransaction.swift
alanpaulin: I like the name `isIdempotentlyEqual` or `isIdempotentlyEquivalent`

the-real-adammork: I was thinking to add "with" so it would read `isIdempotent(with: lhs)` call-site:

```swift
payload.isIdempotent(with: otherPayload)
```

I think the equal qualifier is redundant with idempotent. Not sure the static functions gain much from a rename.

Sources/MobileCoinClient+Deprecated.swift
the-real-adammork: ya, it was a public API for a small amount of time so should be deprecated before removal.

Sources/Transaction/Outputs/PendingTransaction.swift
the-real-adammork: changed to `isIdempotent(with: lhs)`

the-real-adammork:
will merge when CI passes

Sources/Transaction/Outputs/PendingTransaction.swift
the-real-adammork: `isSomething` implies test for equality to "something", `isNotSomething` is the reverse.

the-real-adammork:
putting the merge on hold, until I finish discussing with Bernie.

the-real-adammork:
good for now.

the-real-adammork:
The requirements for Transaction Idempotence have changed, and now only require a Seed to be passed down into the Transaction Builder.

* Add RngSeed type so consumers can pass in a seed to the Transaction Builder
* Remove RNG seedability for txo selection
* Update unit tests
* Add new comparison utils for testing if TxOuts are idempotent (same txOutPublicKey)

PR #197, opened 2022-09-23
Update APIs to use MobileCoinRng instead of RngSeed

carybakker:
Soundtrack of this PR: [A 'random' InSync song lol](https://www.youtube.com/watch?v=6thmPrTxBtI)

### Motivation

Sync up iOS and Android's implementation of idempotence for cross-platform compatibility

### In this PR
* Deprecate the API calls for prepareTransaction that use RngSeed
* De-Deprecate the API calls that use MobileCoinRng
* Deprecated APIs will generate a new RNG from the seed, and then call the de-deprecated APIs with that RNG (all that part is done by the API client code on Android).  Then for equivalency, when we receive an Rng via the de-deprecated APIs (that match Android in that they accept an Rng), we will generate a seed used to create a new ChaCha20Rng for the building of the output transaction (which will match Android once Android is updated to generate the seed before any other use).

### DIFF
--- a/Sources/MobileCoinClient.swift
+++ b/Sources/MobileCoinClient.swift
@@ -203,7 +203,7 @@ public final class MobileCoinClient {
             memoType: memoType,
             amount: amount,
             fee: fee,
-            rngSeed: RngSeed(),
+            rng: MobileCoinChaCha20Rng(),
             completion: completion)
     }
 
@@ -212,11 +212,16 @@ public final class MobileCoinClient {
         memoType: MemoType = .recoverable,
         amount: Amount,
         fee: UInt64,
-        rngSeed: RngSeed,
+        rng: MobileCoinRng,
         completion: @escaping (
             Result<PendingSinglePayloadTransaction, TransactionPreparationError>
         ) -> Void
     ) {
+        guard let rngSeed = rng.generateRngSeed() else {
+            completion(.failure(
+                TransactionPreparationError.invalidInput("Could not create 32 byte RNG seed")))
+            return
+        }
         Account.TransactionOperations(
             account: accountLock,
             fogMerkleProofService: serviceProvider.fogMerkleProofService,
@@ -252,7 +257,7 @@ public final class MobileCoinClient {
             memoType: memoType,
             amount: amount,
             feeLevel: feeLevel,
-            rngSeed: RngSeed(),
+            rng: MobileCoinChaCha20Rng(),
             completion: completion)
     }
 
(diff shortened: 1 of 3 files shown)


voloshyn:
would be great to have a test for this, @dolanbernard can you provide a test data from Android to cross-test?

Sources/MobileCoinClient.swift
```diff
@@ -212,11 +212,16 @@ public final class MobileCoinClient {
         memoType: MemoType = .recoverable,
         amount: Amount,
         fee: UInt64,
-        rngSeed: RngSeed,
+        rng: MobileCoinRng,
         completion: @escaping (
             Result<PendingSinglePayloadTransaction, TransactionPreparationError>
         ) -> Void
     ) {
+        guard let rngSeed = rng.generateRngSeed() else {
```
the-real-adammork, line 220: I dont see this being used below

the-real-adammork: nevermind, needed to expand

the-real-adammork, line 220: not sure this is correct, if you only use the RNG to create a seed which you then pass around to see our internal RNG, that defeats the purpose of letting them pass in their own RNG class

Sources/MobileCoinClient.swift
```diff
@@ -261,11 +266,16 @@ public final class MobileCoinClient {
         memoType: MemoType = .recoverable,
         amount: Amount,
         feeLevel: FeeLevel = .minimum,
-        rngSeed: RngSeed,
+        rng: MobileCoinRng,
         completion: @escaping (
             Result<PendingSinglePayloadTransaction, TransactionPreparationError>
         ) -> Void
     ) {
+        guard let rngSeed = rng.generateRngSeed() else {
```
the-real-adammork, line 274: same feedback here

Sources/MobileCoinClient.swift
carybakker: Android uses the input Rng to seed the RNG for output idempotence.   So here, we are doing the same as Android with respect to the output transaction creation.

Sources/MobileCoinClient.swift
carybakker: same answer as above :)

carybakker:
> would be great to have a test for this, @dolanbernard can you provide a test data from Android to cross-test?

Test has been added with matching data (passing successfully)

Sources/MobileCoinClient.swift
the-real-adammork: well then, doesnt seem to be as flexible as I initially thought, but if its the same as android thats correct

the-real-adammork (approved):

dolanbernard (approved):

PR #205, opened 2022-11-05
Expose "address hash" with a public computed variable on PublicAddress

the-real-adammork:
Soundtrack of this PR: [Formally Unknown - Lotus Turbo [FAM07]](https://www.youtube.com/watch?v=Y4aLpV6SlEk)

### Motivation

Expose the "address hash" publicly. 

### In this PR
* computed var for the public address hash

### DIFF
--- a/Sources/Account/PublicAddress.swift
+++ b/Sources/Account/PublicAddress.swift
@@ -57,6 +57,8 @@ public struct PublicAddress {
     public var spendPublicKey: Data { spendPublicKeyTyped.data }
 
     public var fogReportUrlString: String? { fogInfo?.reportUrlString }
+    
+    public var addressHash: String? { calculateAddressHash()?.hex }
 
     var fogReportUrl: FogUrl? { fogInfo?.reportUrl }
     var fogReportId: String? { fogInfo?.reportId }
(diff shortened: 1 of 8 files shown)


Sources/Account/PublicAddress.swift
```diff
@@ -58,6 +58,8 @@ public struct PublicAddress {
 
     public var fogReportUrlString: String? { fogInfo?.reportUrlString }
 
+    public var addressHash: String? { calculateAddressHash()?.hex }
```
remoun: Let's name this to reflect that it's hex-encoded. We might want a deserialized or base64-encoded variant, to match the Android SDK. 
```suggestion
    public var addressHashHex: String? { calculateAddressHash()?.hex }
```

remoun (approved):
LGTM

Sources/Account/PublicAddress.swift
the-real-adammork: changed to a `Data` object then you can choose hex or base64 like below:

```swift
publicAddress.addressHash?.hexEncodedString()
publicAddress.addressHash?.base64EncodedString()
```

remoun: Beautiful, thank you!

PR #207, opened 2022-11-15
Signed Contingent Inputs

carybakker:
### Motivation

Add support for Signed Contingent Inputs

### In this PR
* create SCI
* prepare and submit tx for SCI
* cancel SCI

### Future Work
* ability to reserve input used by SCI to allow other transactions to be completed without potentially invalidating SCI


Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,162 @@
...
+                    "amountToSpend: \(redacting: amountToSend), " +
+                    "amountToRecieve: \(redacting: amountToReceive), " +
+                    "unspentTxOutValues: \(redacting: unspentTxOuts.map { $0.value })",
+                logFunction: false)
+
+            // fee is zero here, because the fee will be covered by the consumer of the SCI
+            switch txOutSelector
+                .selectTransactionInputs(amount: amountToSend, fee: 0, fromTxOuts: unspentTxOuts)
+                .mapError({ error -> TransactionPreparationError in
+                    switch error {
```
the-real-adammork: do you need the re-map the error here if you create a new one in the failure case ?

Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,162 @@
...
+                    case .insufficientTxOuts:
+                        return .insufficientBalance()
+                    case .defragmentationRequired:
+                        return .defragmentationRequired()
+                    }
+                })
+            {
+            case .success(let txOutsToSpend):
+                metaFetcher.blockVersion {
```
the-real-adammork, line 129: this case is too long. I think we could tighten it up by moving some things into a convenience functions

Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,162 @@
...
+                    switch $0 {
+                    case .success(let blockVersion):
+
+                        // verify block version >= 3
+                        guard blockVersion >= 3 else {
+                            serialQueue.async {
+                                completion(.failure(.invalidBlockVersion(
```
the-real-adammork:
```suggestion
                                completion(.failure(.requiresBlockVersion3))
```

```swift
fileprivate extension SignedContingentInputCreationError {
    static var requiresBlockVersion3: SignedContingentInputCreationError = .invalidBlockVersion("Block version must be > 3 for SCI support")
}
```

Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,162 @@
...
+
+                        signedContingentInputCreator.createSignedContingentInput(
+                            inputs: txOutsToSpend,
+                            recipient: recipient,
+                            memoType: memoType,
+                            amountToSend: amountToSend,
+                            amountToReceive: amountToReceive,
+                            tombstoneBlockIndex: tombstoneBlockIndex,
+                            blockVersion: blockVersion) { result in
+                                completion(result.mapError {
```
the-real-adammork: maybe do:

```swift
fileprivate extension SignedContingentInputCreationError {
    static func make(_ transactionPreperationError: TransactionPreperationError) {
        switch $0 {
        case .invalidInput(let reason):
            error = .invalidInput(reason)
        case .defragmentationRequired(let reason):
            error = .defragmentationRequired(reason)
        case .insufficientBalance(let reason):
            error = .insufficientBalance(reason)
        case .connectionError(let reason):
            error = .connectionError(reason)
        }
    }
}

```

then you can compact to something like: 

```swift
completion(result.mapError { error -> SignedContingentInputCreationError in
    .make(error)
}
```

or 

```swift
completion(result.mapError {
    SignedContingentInputCreationError.make($0)
}
```

Sources/Account/Account+TransactionOperations.swift
```diff
@@ -281,5 +281,99 @@ extension Account {
...
+                            switch error {
+                            case .insufficientTxOuts:
+                                return .insufficientBalance()
+                            case .defragmentationRequired:
+                                return .defragmentationRequired()
+                            }
+                        })
+                    {
+                    case .success(let (inputs: inputs, fee: fee)):
+                        metaFetcher.blockVersion {
```
the-real-adammork, line 448: you can get both blockversion & fee at the same time:

```swift
func fetchMeta(completion: @escaping (Result<MetaCache, ConnectionError>) -> Void) { ... }
```

Sources/Account/Account+TransactionOperations.swift
```diff
@@ -281,5 +281,99 @@ extension Account {
...
+                        return
+                    }
+
+                    logger.info(
+                        "Preparing pre-signed input transaction with fee level...feeLevel: " +
+                            "\(feeLevel), unspentTxOutValues: " +
+                            "\(redacting: unspentTxOuts.map { $0.value })",
+                        logFunction: false)
+
+                    switch self.txOutSelector
```
the-real-adammork, line 433: nitpick: this statement is too long for a switch case, id put it on its own line and switch on its result, or remove the mapError part, similar to my suggestion on the previous file.

Sources/Account/Account+TransactionOperations.swift
```diff
@@ -281,5 +281,99 @@ extension Account {
...
+                                self.transactionPreparer.preparePresignedInputTransaction(
+                                    presignedInput: presignedInput,
+                                    inputs: inputs,
+                                    memoType: memoType,
+                                    amount: amountToSend,
+                                    fee: Amount(fee, in: presignedInput.feeTokenId),
+                                    tombstoneBlockIndex: tombstoneBlockIndex,
+                                    blockVersion: blockVersion,
+                                    completion: completion)
+                            case .failure(let error):
```
the-real-adammork, line 468: nitpick: You could convert the fee & blockversion checks to `guards` to avoid some of these nested .failure cases.

Sources/Transaction/SignedContingentInput.swift
```diff
@@ -0,0 +1,96 @@
...
+
+        self = signedContingentInput
+    }
+
+    public var serializedData: Data {
+        proto.serializedDataInfallible
+    }
+
+    public var isValid: Bool {
+        true
```
the-real-adammork:
```suggestion
        true // TODO - debug
```

Sources/Transaction/TransactionBuilder.swift
```diff
@@ -30,34 +30,49 @@ extension TransactionBuilderError: CustomStringConvertible {
...
 
 final class TransactionBuilder {
+
     private let tombstoneBlockIndex: UInt64
 
     private let ptr: OpaquePointer
 
     private let memoBuilder: TxOutMemoBuilder
 
+    internal struct Context {
```
the-real-adammork: internal is the default access modifier

the-real-adammork: glad you did this tho ! simplifies these a TON

Sources/Transaction/TransactionBuilder.swift
```diff
@@ -30,34 +30,49 @@ extension TransactionBuilderError: CustomStringConvertible {
...
+        let fee: Amount
+    }
+
     private init(
-        fee: Amount,
-        tombstoneBlockIndex: UInt64,
-        fogResolver: FogResolver = FogResolver(),
-        memoBuilder: TxOutMemoBuilder = DefaultMemoBuilder(),
-        blockVersion: BlockVersion
+        context: InnerContext
```
the-real-adammork, line 59: 🦾

the-real-adammork:
finished my initial pass, and other than a few nits this looks solid !

Tests/Integration/Network/NetworkConfigFixtures.swift
```diff
@@ -9,8 +9,8 @@ import Foundation
 import XCTest
 
 enum NetworkConfigFixtures {
-    static let alphaDev = DynamicNetworkConfig.AlphaDevelopment.make()
-    static let network: NetworkPreset = .testNet
+    // static let alphaDev = DynamicNetworkConfig.AlphaDevelopment.make()
```
the-real-adammork: necc. ?

Sources/Account/Account+SCIOperations.swift
carybakker: fixed

Sources/Account/Account+SCIOperations.swift
carybakker: pulled a few things out to private helper functions

Tests/Integration/Network/NetworkConfigFixtures.swift
carybakker: restored original

Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,169 @@
...
+            amountToSend: Amount,
+            amountToReceive: Amount,
+            completion: @escaping (
+                Result<SignedContingentInput, SignedContingentInputCreationError>
+            ) -> Void
+        ) {
+            let functionName = "createSignedContingentInput"
+
+            guard verifyAmountIsNonZero(amountToSend, "send", completion) else {
+                return
```
the-real-adammork, line 105: these `guards` should return Result - errors

the-real-adammork: oh I see, they take a completion.

Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,169 @@
...
+        func createSignedContingentInput(
+            to recipient: PublicAddress,
+            memoType: MemoType,
+            amountToSend: Amount,
+            amountToReceive: Amount,
+            completion: @escaping (
+                Result<SignedContingentInput, SignedContingentInputCreationError>
+            ) -> Void
+        ) {
+            let functionName = "createSignedContingentInput"
```
the-real-adammork: this can be done automatically by the logger - https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/master/Sources/Common/MobileCoinLogging.swift

Sources/MobileCoinClient.swift
```diff
@@ -189,6 +189,40 @@ public final class MobileCoinClient {
...
+        completion: @escaping (
+            Result<SignedContingentInput, SignedContingentInputCreationError>
+        ) -> Void
+    ) {
+        guard let rngSeed = defaultRng.generateRngSeed() else {
+            completion(.failure(
+                SignedContingentInputCreationError.invalidInput("Couldn't create 32byte RNG seed")))
+            return
+        }
+        Account.SCIOperations(
```
the-real-adammork, line 205: I doesn't make sense to init this struct and not store it anywhere. the init part should be moved into `Account`
s init then call "SCIOperation"s on that variable

Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,169 @@
...
+            guard verifyAmountIsNonZero(amountToReceive, "receive", completion) else {
+                return
+            }
+
+            // get all unspent txOuts
+            let (unspentTxOuts, ledgerBlockCount) =
+            account.readSync {
+                ($0.unspentTxOuts(tokenId: amountToSend.tokenId), $0.knowableBlockCount)
+            }
+            logUnspentTxOuts(recipient, amountToSend, amountToReceive, unspentTxOuts)
```
the-real-adammork, line 123: This feels like debug code, so it shouldn't be merged to master.

Sources/Account/Account+SCIOperations.swift
```diff
@@ -0,0 +1,169 @@
...
+                logFunction: false)
+        }
+
+        private func logUnspentTxOuts(
+            _ recipient: PublicAddress,
+            _ amountToSend: Amount,
+            _ amountToReceive: Amount,
+            _ txOuts: [KnownTxOut]
+        ) {
+            logger.info(
```
the-real-adammork: nitpick: since these functions don't use any member variables can we move them to the logger file `MobileCoinLogging` ?

Sources/Account/Account+TransactionOperations.swift
```diff
@@ -281,5 +281,214 @@ extension Account {
...
+            let amountToReceive = presignedInput.rewardAmount
+
+            let (unspentTxOuts, ledgerBlockCount) =
+                self.account.readSync {
+                    ($0.unspentTxOuts(tokenId: amountToSend.tokenId),
+                     $0.knowableBlockCount)
+                }
+
+            guard ledgerBlockCount <= presignedInput.tombstoneBlockIndex else {
+                serialQueue.async {
```
the-real-adammork, line 302: fwiw I prefer this approach to guards over the one where its factored into a function in `Sources/Account/Account+SCIOperations.swift`

Sources/Account/Account+SCIOperations.swift
the-real-adammork: nitpick: looking at the function, some of the logging could be useful, Id inline it instead of having it as a function

Sources/Account/Account+SCIOperations.swift
the-real-adammork: nitpick: would prefer these verify functions are inlined

Sources/MobileCoinClient.swift
the-real-adammork: although I see this is an existing pattern so we can keep it the way you have it now.

Sources/Transaction/SignedContingentInputBuilderUtils.swift
```diff
@@ -0,0 +1,213 @@
...
+                        publicAddressPtr,
+                        rngCallbackPtr,
+                        confirmationNumberPtr,
+                        &errorPtr)
+                }).mapError {
+                    switch $0.errorCode {
+                    case .invalidInput:
+                        return .invalidInput("\(redacting: $0.description)")
+                    case .attestationVerificationFailed:
+                        return .attestationVerificationFailed("\(redacting: $0.description)")
```
the-real-adammork: commented on the other PR, but I don't think this function can throw an attestation error

Sources/Transaction/SignedContingentInputBuilderUtils.swift
```diff
@@ -0,0 +1,213 @@
...
+                            amount.tokenId.value,
+                            rngCallbackPtr,
+                            confirmationNumberPtr,
+                            &errorPtr)
+                    }).mapError {
+                        switch $0.errorCode {
+                        case .invalidInput:
+                            return .invalidInput("\(redacting: $0.description)")
+                        case .attestationVerificationFailed:
+                            return .attestationVerificationFailed(
```
the-real-adammork: same here, not sure if this can be thrown

Sources/Transaction/SignedContingentInputCreator.swift
```diff
@@ -0,0 +1,180 @@
...
+                forRealTxOutIndices: inputs.map { $0.globalIndex },
+                selectionRange: ledgerTxOutCount.map { ..<$0 }
+            ).map { Array($0) }
+
+        // There's a chance that a txo we selected as a mixin is in a block that's greater than
+        // the highest block of our inputs, in which case, using the highest block of our inputs
+        // as the requested merkleRootBlock might cause getOutputs to fail. However, the current
+        // implementation of the server doesn't use the merkleRootBlock param, so for the time
+        // being, we will ignore this concern.
+        let merkleRootBlock = merkleRootBlock ?? inputs.map { $0.block.index }.reduce(0, max)
```
the-real-adammork, line 126: good comment

the-real-adammork (changes requested):
A few outstanding questions that need to be resolved before we merge.

Then for "nitpicks:" you can choose which ones you want todo.

Sources/Transaction/SignedContingentInputCreator.swift
carybakker: Copied from existing transaction prep code 👍

Sources/Transaction/SignedContingentInputBuilderUtils.swift
carybakker: removed

Sources/Transaction/SignedContingentInputBuilderUtils.swift
carybakker: removed attestationVerificationFailed case

Sources/Account/Account+SCIOperations.swift
carybakker: awesome - tx

Sources/Account/Account+SCIOperations.swift
carybakker: Seems like MobileCoinLogging is much more generic / reusable / general purpose...  

This was code that was inline originally (starting code taken from Account+TransactionOperations.swift, where it's inlined there)... not sure why I pulled it out, maybe was just trying to simplify the body.

I re-inlined it for now, but I don't think that MobileCoinLogging is the right place... maybe someplace else if you don't like the inlining - like an extension at bottom or in another related filename for A+SCOOps

Sources/Account/Account+SCIOperations.swift
carybakker: Ah, ya, I don't recall for sure, but I explicitly disabled the logging of the function name in the default manner to be able to log "<function name> failure" in the log... But sure, no need to do that - can use the default.  Switched it out.

Sources/Account/Account+SCIOperations.swift
carybakker: 👍 inlined...

Sources/Account/Account+TransactionOperations.swift
carybakker: ✅

Sources/Account/Account+TransactionOperations.swift
carybakker: will revisit nits after merge

Sources/Account/Account+TransactionOperations.swift
carybakker: will revisit nits after merge

Sources/Account/Account+SCIOperations.swift
carybakker: will revisit nits after merge

the-real-adammork:
whats left on this PR ?

carybakker:
> whats left on this PR ?

Should be good to go

the-real-adammork (approved):
all looks good

PR #214, opened 2022-12-07
Update to support v1 and v2 of masked amounts and masked token ids

carybakker:
### Motivation

The mobilecoin lib was updated to have multiple versions of masked amount and masked token id.  This change is to update the Swift SDK to support the the masked amount version changes in mobilecion and libmobilecoin.

### In this PR
* support for masked amount and masked token id v1 and v2

### DIFF
--- a/Sources/Fog/View/FogView.swift
+++ b/Sources/Fog/View/FogView.swift
@@ -193,6 +193,14 @@ extension LedgerTxOut {
         -> Result<LedgerTxOut, ConnectionError>
     {
         guard let ledgerTxOut = LedgerTxOut(txOutRecord, viewKey: viewKey) else {
+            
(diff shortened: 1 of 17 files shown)


Sources/Fog/View/FogView.swift
```diff
@@ -193,6 +193,14 @@ extension LedgerTxOut {
         -> Result<LedgerTxOut, ConnectionError>
     {
         guard let ledgerTxOut = LedgerTxOut(txOutRecord, viewKey: viewKey) else {
+
```
carybakker, line 196: @the-real-adammork did you mean to double up the `guard let`s?

the-real-adammork: no that was for testing

PR #218, opened 2022-12-16
Change Memo encoding to include typeBytes, typeName and "data"

the-real-adammork:
### Motivation

Change encoding format, update unit tests, remove typeBytes enum add as static vars to memo types.

### In this PR
* Update encoding format

### Future Work
* make protobuf interface layer for all APIs so they can be consistent between android and iOS


voloshyn (approved):
lgtm

PR #219, opened 2022-12-16
Update `printBalances`, change "wait for expectations" behavior 

carybakker:
### Motivation

Protocol tests were having issues where the waitForExpectation for one protocol test was being fulfilled incorrectly by a prior protocol test.  Also, the protocol tests seemed to include a sleep between testing the protocols, presumably to try to prevent this issue?  These changes also remove the need to sleep, which will reduce each test execution by 10s.

### In this PR
    
     * Previously, when an expect would time out, the code would sleep for
       10s and attempt to test the subsequent protocol.  However when the
       previous protocol eventually fulfills the expectation, it will
       confuse the next test as the current code waits for any expectation
    
     * This fix updates the protocol test loop to wait for the specific
       expectation for the protocol to prevent the multiple expectation
       fullfillment error.
    
     * Additionally, this update speeds up the protocol tests as there
       should no longer be any need to sleep for 10s between protocols
       (I believe this was an attempt to prevent the above noted error from
       happening by giving time for the prior protocol to fulfill the
       expectation after timeout possibly?  not sure)### Future Work

     * Includes another small fix to remove magic number around number
       test accounts when printing balances

### DIFF
--- a/Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
+++ b/Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
@@ -21,7 +21,7 @@ extension XCTestCase {
             let description = "[\(transportProtocol.description)]:\(description)"
             print("Testing ... \(description)")
             let expect = expectation(description: description)
-            try testCase(transportProtocol, expect) 
+            try testCase(transportProtocol, expect)
             XCTWaiter().wait(for: [expect], timeout: timeout)
 
             // Without the sleep here, we were occasionally seeing 'insufficientFunds' errors
--- a/Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
+++ b/Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
@@ -16,12 +16,17 @@ extension XCTestCase {
     ) rethrows {
         let supportedProtocols = TransportProtocol.supportedProtocols
 
+        let last = supportedProtocols.last
         try supportedProtocols.forEach { transportProtocol in
             let description = "[\(transportProtocol.description)]:\(description)"
             print("Testing ... \(description)")
             let expect = expectation(description: description)
             try testCase(transportProtocol, expect)
             XCTWaiter().wait(for: [expect], timeout: timeout)
+
+            // Without the sleep here, we were occasionally seeing 'insufficientFunds' errors
+            // caused by tests that transact, where the change txo has not come back yet
+            sleep(transportProtocol == last ? 0 : interval)
         }
     }
 
--- a/Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
+++ b/Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
@@ -15,14 +15,13 @@ extension XCTestCase {
                 _ testCase: (TransportProtocol, XCTestExpectation) throws -> Void
     ) rethrows {
         let supportedProtocols = TransportProtocol.supportedProtocols
-        let last = supportedProtocols.last
+
         try supportedProtocols.forEach { transportProtocol in
             let description = "[\(transportProtocol.description)]:\(description)"
             print("Testing ... \(description)")
             let expect = expectation(description: description)
             try testCase(transportProtocol, expect)
-            waitForExpectations(timeout: timeout)
-            sleep(transportProtocol == last ? 0 : interval)
+            XCTWaiter().wait(for: [expect], timeout: timeout)
         }
     }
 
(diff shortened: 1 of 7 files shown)


the-real-adammork:
we need some sleep time in between integration tests, otherwise the TxOuts will be in-flight from a previous test when we need them for something else.

carybakker:
> we need some sleep time in between integration tests, otherwise the TxOuts will be in-flight from a previous test when we need them for something else.

Ah, ok.. I'll update to include the ability to sleep for tests that need it.

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
```diff
@@ -45,27 +45,28 @@ class MobileCoinClientPublicApiIntTests: XCTestCase {
...
                     print("account index \(index) public address \(key.publicAddress)")
                     print("account index \(index) balance: \(amountPicoMob)")
                 }
-
-                if index == 9 {
-                    expect.fulfill()
-                }
             }
+
+            XCTWaiter().wait(for: [balExpect], timeout: 40)
```
the-real-adammork: this doesn't look correct, I think you want to create an array of expectations, then pass them all in to XCTWaiter

the-real-adammork:
only one `waitForExpectations` can be active at once, so Im not sure your point about "previous test expectation succeeding" is correct - https://developer.apple.com/documentation/xctest/xctestcase/1500748-waitforexpectations

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
carybakker: This will serialize the protocol tests... should be identical to the prior code, with the exception that a delayed expectation fulfillment from, say, http... can't clobber the grpc iteration (the error I was seeing)

Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
```diff
@@ -15,13 +15,17 @@ extension XCTestCase {
...
         let supportedProtocols = TransportProtocol.supportedProtocols
+
         let last = supportedProtocols.last
         try supportedProtocols.forEach { transportProtocol in
             let description = "[\(transportProtocol.description)]:\(description)"
             print("Testing ... \(description)")
             let expect = expectation(description: description)
             try testCase(transportProtocol, expect)
-            waitForExpectations(timeout: timeout)
+            XCTWaiter().wait(for: [expect], timeout: timeout)
```
the-real-adammork, line 25: this is not any different, it still waits on one expectation

carybakker: But it doesn't care which one...   Here's what I was seeing:

the-real-adammork (changes requested):
the changes to `Tests/Common/Utils/XCTestCase+SupportedProtocols.swift` do not do anything, would prefer we keep that as-is for now and refactor when we do the integration-test speed-up tech debt task.

The printBalances() test can and should be improved, if you loop over all the accounts and create an array of expectations the XCTWaiter() looks like the correct solution there.

Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
carybakker: * http test runs
* wait for expectation for http test times out (b/c server n/a), but http process still in flight
* grpc test runs
* http test finally fulfills expectation due to failure
* grpc test fails, fulfilling expectation

carybakker:
> the changes to `Tests/Common/Utils/XCTestCase+SupportedProtocols.swift` do not do anything, would prefer we keep that as-is for now and refactor when we do the integration-test speed-up tech debt task.
> 
> The printBalances() test can and should be improved, if you loop over all the accounts and create an array of expectations the XCTWaiter() looks like the correct solution there.

The fix an existing error when servers are down... it's reproducible with master branch, and fixed by these changes.

carybakker:
> only one `waitForExpectations` can be active at once, so Im not sure your point about "previous test expectation succeeding" is correct - https://developer.apple.com/documentation/xctest/xctestcase/1500748-waitforexpectations

Will double check this, because it seems to work as expected, and I'm technically not using waitForExpectations... Given that the code is working, I'm assuming that it may be ok to use XCTWaiter() in parallel.. will verify

Tests/Common/Utils/XCTestCase+SupportedProtocols.swift
the-real-adammork: Apple recommends being explicit in the expectations being waited for - https://developer.apple.com/documentation/xctest/xctestcase/1500748-waitforexpectations

the-real-adammork: resolving

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
the-real-adammork: the code is creating ~9 balances checks, we should wait for all of them to complete using the array of expectations passed to XCTWaiter()

the-real-adammork (approved):

PR #220, opened 2023-01-10
Update LICENSE

collinjackson:
### In this PR

* Updates the Swift SDK license from GPLv3 to Apache 2.


the-real-adammork (approved):
cool. I updated the version number so CICD makes a new release after merge.

PR #228, opened 2023-02-06
Remove duplicate tests already moved to async

carybakker:
Soundtrack of this PR: [Clean Up](https://www.youtube.com/watch?v=oY-H2WGThc8)

### Motivation

Remove redundant tests to speed up integration testing

### In this PR
* Removed non-async tests already implemented in asyc form
* Add logic to pre-populate transactions with all MemoTypes for RTH testing
* Converted remaining non-async tests async

[Fixes #184383403]

### Future Work
* Continued removal of test redundancy


Tests/Integration/MobileCoinClientPublicApiIntTests.swift
```diff
@@ -1,538 +1,12 @@
...
-                checkBalances(client, expect) { balancesBefore in
-                    prepareAndSubmit(client, expect, fee) {
-                        verifyBalanceChange(client, balancesBefore, expect)
-                    }
-                }
-            }
-        }
-    }
-
-    func testCreateSignedContingentInput() throws {
```
the-real-adammork, line 428: I dont see this in the async tests

Tests/Integration/MobileCoinClientPublicAsyncApiIntTests.swift
```diff
@@ -14,24 +14,6 @@ import XCTest
 @available(iOS 15.0, *)
 class MobileCoinClientPublicAsyncApiIntTests: XCTestCase {
 
-    func testBalance() async throws {
```
the-real-adammork, line 17: would prefer to keep this test and the balances test. the Print balances test is for debugging and will never fail

the-real-adammork: we can `XCTSkip(true)` the printBalances test, and then we can use it when we need it by commenting that line in debugging

carybakker: Since we have them available in async form for debugging, any reason why we need the non-async versions???

We'll get the same info...

carybakker: Sure, but just not sure why we need it... if it's not testing anything, we can just use the async version.  Any reason to keep and maintain two versions of the functionality to just print balances?

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
carybakker: Ah, right... in the first pass I had left this off because the create is included in the submit SCI test (i.e.  an SCI needs to be created to submit).

I'll get the RTH one passing with the test data, migrate the RTH and create SCI tests to async, and make a note here when done.

Tests/Integration/MobileCoinClientPublicAsyncApiIntTests.swift
carybakker: Oh, man... sorry - I just realized this is from the Async call, and you were referring to the removal of the single 'testBalance' async test.

carybakker: will put this one back and skip...

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
carybakker: * RTH Test Data was updated manually at first
* Moved RTH test to Async
* Logic added to attempt a pre-population of RTH test data (MemoTypes), test will fail if all types not subsequently present
* Renamed MobileCoinClientPublicAsyncApiIntTests.swift -> MobileCoinClientPublicApiIntTests.swift (i.e. removed "Async") as the previous MobileCoinClientPublicApiIntTests.swift was empty

Tests/Integration/MobileCoinClientPublicAsyncApiIntTests.swift
carybakker: added `testBalance` and `testPrintBalances` back, have them both set to XCTSkip() to save integration test execution time (balances are logged when updated for the tests as part of the update balance logic in any case)

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
```diff
@@ -1,573 +1,821 @@
...
             }
 
-            balanceExpects.append(expect)
-        })
-
-        XCTWaiter().wait(for: balanceExpects, timeout: 40)
-        expect.fulfill()
-    }
-
-    func testBalances() throws {
```
the-real-adammork, line 83: did this test make it into your additions ?

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
```diff
@@ -1,573 +1,821 @@
...
+                    try await Task.sleep(nanoseconds: UInt64(2 * 1_000_000_000))
+                    try await checkBalance()
+                    return
+                }
+            } catch {}
         }
+        try await checkBalance()
     }
 
-    func createSignedContingentInput(
```
the-real-adammork, line 437: also this one ?

carybakker: Just added an async version of this back in.

Also, updated the logic for the sci tests to be enabled for testNet in addition to Alpha in Network Presets now that TestNet is on block version 3.

Tests/Integration/MobileCoinClientPublicApiIntTests.swift
carybakker: Added an async version back in - tx

the-real-adammork (approved):

PR #230, opened 2023-02-08
Use 14.1 for Workflows that don't depend on GRPC

the-real-adammork:
Soundtrack of this PR: [Anz - Inna Step](https://soundcloud.com/yung-anz/inna-circle?in=ninja-tune/sets/anz-you-could-be-ft-george)

### Motivation

GRPC dependencies need Xcode 12.5.1 to compile, so setting 14.1 for workflows that don't require that to increase our coverage.

### In this PR
* use Xcode 14.1 for workflows if possible

### Future Work
* Find long-term fix for compilation issue by contacting cocoapods, or switching to SPM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 defaults:
-  default-xcode-version: &default-xcode-version "12.5.1"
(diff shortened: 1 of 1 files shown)


carybakker (approved):
LGTM

PR #231, opened 2023-02-16
Cary/service url rotation unit tests 184441012

carybakker:
Soundtrack of this PR: [Converting Units](https://www.youtube.com/watch?v=wldRifX1Ico)

### Motivation

A number of tests that are implemented as Integration Tests should really be handled via Unit Tests.  Integration Tests will still handle the holistic testing of functionality, but granular tests for code coverage or specific cases can be more efficiently handled with Unit Tests. 

This has become much more important as the amount of time required to execute Integration Tests has increased with the number of Integration Tests implemented, to the point where it impacts out ability to release in a timely manner.

### In this PR
 * This PR moves a few of the test candidates for this change, around URL Load balancing, to Unit Tests to improve Integration Test execution time.

### Future Work
* additional integration tests for subsets of full flows or specific cases for code coverage may also be moved to unit tests in the future


Tests/Fixtures/UnitTestFixtures.swift
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    static func createFogViewConnection(
+        using fogUrlLoadBalancer: SequentialUrlLoadBalancer<FogUrl>
+    ) throws -> FogViewConnection {
+        let networkConfig = try failingHttpNetworkConfig(
+            fogUrlLoadBalancer,
+            UrlLoadBalancerFixtures().invalidConsensusUrlBalancer)
+
+        fogUrlLoadBalancer.rotationEnabled = false
```
the-real-adammork, line 47: why do you set this to false here and then true later ?

Tests/Fixtures/UrlLoadBalancerFixtures.swift
```diff
@@ -0,0 +1,270 @@
...
+        let urlTypeName = "Consensus"
+        let seqLoadBalancer: SequentialUrlLoadBalancer<ConsensusUrl>
+        let blockchain: BlockchainConnection
+
+        var loadBalancer: MockUrlLoadBalancer {
+            seqLoadBalancer
+        }
+
+        init() throws {
+            self.seqLoadBalancer = try UrlLoadBalancerFixtures().invalidConsensusUrlBalancer
```
the-real-adammork: This PR adds alot of abstractions that could be hard to GROK, can you collapse these structs inline where the test is called ? Will improve readability and its essentially the test-code so its ok for it to be a little longer at the call-site.

Tests/Fixtures/UnitTestFixtures.swift
```diff
@@ -0,0 +1,162 @@
...
+            fogReport: attestation)
+
+        var networkConfig = try NetworkConfig.make(
+            consensusUrlLoadBalancer: consensusUrlLoadBalancer,
+            fogUrlLoadBalancer: fogUrlLoadBalancer,
+            attestation: attestationConfig,
+            transportProtocol: .http)
+            .get()
+
+        networkConfig.httpRequester = MockHttpRequester()
```
the-real-adammork: can you dependency inject this MockHttpRequester, will make the test call-site more readable cus people will be able to see that we're using a MockHttpRequester which will explain its unit-test like behavior

Tests/Mocks/MockHttpRequester.swift
```diff
@@ -0,0 +1,82 @@
+//
+//  Copyright (c) 2020-2023 MobileCoin. All rights reserved.
+//
+@testable import MobileCoin
+import XCTest
+
+public class MockHttpRequester: NSObject, HttpRequester {
```
the-real-adammork: The "always fail" behavior of this mock class is not captured in the name or in its init pattern. I would change the name, or add an init pattern so when you create the MockHttpRequester you can choose the `Result<HTTPResponse, Error>` that you want mocked on request responses.

the-real-adammork (changes requested):
remove/collapse this file `Tests/Fixtures/UrlLoadBalancerFixtures.swift` in-line where the tests are called, plus some other small changes

Tests/Fixtures/UnitTestFixtures.swift
carybakker: The initial creation of the service was rotating as a side effect during the creation - so to make sure that we're detecting rotation as a result of the attempted call, the rotation is disabled during service creation, and enabled for the test call (this only applies to testing)

Tests/Fixtures/UrlLoadBalancerFixtures.swift
carybakker: This was the same format used for the prior tests - so that's why I repeated this...

Using the fixtures reduced a lot of redundant code - since the creation of the service couldn't be abstracted/generalized, it was wrapped in a fixture when I wrote the integration tests.

You sure you want to change it?

Tests/Fixtures/UnitTestFixtures.swift
carybakker: done

Tests/Fixtures/UrlLoadBalancerFixtures.swift
carybakker: de-abstracted tests a bit from the fixture so you can see the individual test logic a bit more clearly

Tests/Mocks/MockHttpRequester.swift
carybakker: updated name to MockFailingHttpRequester - good suggestion

carybakker:
> remove/collapse this file `Tests/Fixtures/UrlLoadBalancerFixtures.swift` in-line where the tests are called, plus some other small changes

I didn’t fully collapse the test fixture file (just noticed that comment at the end), but I did put the config code relevant to the tests inline…

I think it does what you were looking for in terms of ‘demonstrating’ the intent of the test without bloating the test file with all the config of the individual service connections

Tests/Fixtures/UrlLoadBalancerFixtures.swift
```diff
@@ -30,17 +29,10 @@ extension UrlLoadBalancerFixtures {
     struct Blockchain: ServiceFixture {
         let serviceName = "Blockchain"
```
the-real-adammork, line 30: where are these strings used ?

the-real-adammork (approved):
If those extra strings in the fixture file can be removed, that would be ideal, if not, ok to merge

Tests/Fixtures/UrlLoadBalancerFixtures.swift
carybakker: Ah, yes... they were vestigial strings from the original code - will remove - thanks!

carybakker: Actually, they're used to provide contextual information for the testing logic in `func testUrlRotation(fixture: ServiceFixture, with transProto: TransportProtocol)`... but this did help me catch an error where I had the URL type set to "Consensus" in that code, instead of setting it to the value for the particular service fixture (the definitions you saw).

PR #237, opened 2023-03-02
Cary/parallelize integration tests 184551676

carybakker:
Soundtrack of this PR: [Seven - Parallel](https://www.youtube.com/watch?v=JdewJT1Yirk)

### Motivation

Speed up Integration test execution in GHA to enable for PRs

### In this PR
* Split up Integration tests into transacting / non-transacting
* enabled parallel setting for test groups
* created separate jobs in workflow for transacting / non-transacting
* manually segmented test account indices to avoid test wallet clobbering by parallel tests
* switch from xcpretty to xcbeautify

### Future Work
* automate the test wallet config for tests with run-specific wallets


Example/Example/Pods-Example.debug.xcconfig
```diff
@@ -0,0 +1,22 @@
+ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES
```
the-real-adammork, line 1: Wondering if we need to check this file in ? I think its generated from cocoapods.

ExampleHTTP/Example.xcworkspace/contents.xcworkspacedata
```diff
@@ -7,7 +7,4 @@
    <FileRef
       location = "group:Pods/Pods.xcodeproj">
    </FileRef>
-   <FileRef
-      location = "group:/Users/adam/Projects/MobileCoin/MobileCoin-Swift/Example/TestPlans/Protocol Unit Tests.xctestplan">
```
the-real-adammork, line 11: do you know why this is being removed ?

the-real-adammork (approved):
two questions, otherwise this looks great ! super excited to get it merged

ExampleHTTP/Example.xcworkspace/contents.xcworkspacedata
the-real-adammork: my bad

PR #242, opened 2023-03-18
Cary/dynamic test wallets 184577127

carybakker:
Soundtrack of this PR: [Never Broke Again](https://www.youtube.com/watch?v=8aYsAaZ8RLM)

### Motivation

Integration test robustness by having distinct wallets per run, avoiding test account collisions, and the ever-persistent low-balance errors from improperly shared accounts

### In this PR
* TestAccountFactory capable of creating arbitrarily configured transactions for new test wallets

fixes #184577127

### Future Work
* [[iOS SDK] Link up dynamic accounts with Integration tests](https://www.pivotaltracker.com/story/show/184740454)


Tests/Util/TestAccountFactory.swift
```diff
@@ -0,0 +1,116 @@
...
+            fogReportId: self.fogReportId,
+            fogAuthoritySpki: self.fogAuthoritySpki)
+            .get()
+
+        for token in testAccountConfig.txData.keys {
+            guard let amounts = testAccountConfig.txData[token] else {
+                throw TestAccountFactoryError.unknownError("no amounts provided")
+            }
+
+            for amount in amounts {
```
the-real-adammork: would be more efficient to sum up the all the amounts and send that as one transaction

Tests/Util/TestAccountFactory.swift
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    let fogReportUrl: String
+    let fogReportId: String = ""
+    let fogAuthoritySpki: Data
+    let accountIndex: UInt32 = 0
+
+    public struct TestAccountConfig {
+        let name: String
+        let txData: [TokenId: [UInt64]]
```
the-real-adammork: can we document what txData is storing ? or maybe instead of storing UInt64, store Amount objects then we're surfacing those semantics to the next person.

Tests/Util/TestAccountFactory.swift
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public struct TestAccount {
```
the-real-adammork, line 44: we could store the config object and make `name` a computed variable then we have one source of truth for it.

Tests/Integration/Transacting/MobileCoinClientInternalIntTests.swift
```diff
@@ -287,4 +287,173 @@ class MobileCoinClientInternalIntTests: XCTestCase {
...
+            TestAccountFactory.TestAccountConfig(
+                name:"acct0",
+                txData:
+                    [.MOB : [1 + minFee]]),
+            TestAccountFactory.TestAccountConfig(
+                name:"acct1",
+                txData:
+                    [
+                        .MOB : [1, 2, minFee],
+                        .MOBUSD : [1 , 2, 3, minMOBUSDFee]
```
the-real-adammork, line 370: if you like the convenience of this initializer w/ integers in the array, you could write a custom init function that converts those UInt64's into amounts, then we get the semantics of "Amount" without the annoyance of writing them each time.

Tests/Integration/Transacting/MobileCoinClientInternalIntTests.swift
```diff
@@ -287,4 +287,173 @@ class MobileCoinClientInternalIntTests: XCTestCase {
...
+    
+#if swift(>=5.5)
+
+    @available(iOS 15.0, *)
+    func testDynamicAccountCreation() async throws {
+        // try XCTSkip()
+        let minFee = IntegrationTestFixtures.fee
+         let minMOBUSDFee: UInt64 = 2650
+
+        func verifyBalances(
```
the-real-adammork:
```suggestion
        func verifyAndReturnBalances(
```

Tests/Util/TestAccountFactory.swift
carybakker: The whole point is to create individual transactions for test purposes - no need to sum them up.  If we want an account with a single bal TxO, that's what we pass in... kwim?

Tests/Util/TestAccountFactory.swift
carybakker: Sure, I can switch this to just an array of Amounts... In fact, that gives more control over the ordering of the transactions created (the way it is now will group the individual transactions by token ID).

Tests/Util/TestAccountFactory.swift
carybakker: Sounds good -

Tests/Integration/Transacting/MobileCoinClientInternalIntTests.swift
carybakker: Great idea

Tests/Integration/Transacting/MobileCoinClientInternalIntTests.swift
carybakker: Hmm.. what do we need to return the balances for here?  This is just the test verification code - making sure all the new account balances match the expected amount totals...

Tests/Util/TestAccountFactory.swift
carybakker: done

Tests/Util/TestAccountFactory.swift
carybakker: done

Tests/Integration/Transacting/MobileCoinClientInternalIntTests.swift
carybakker: done

Tests/Integration/Transacting/MobileCoinClientInternalIntTests.swift
carybakker: Resolving for now, but lmk if you see any real need for this after review of comment - thanks!

the-real-adammork (approved):

PR #243, opened 2023-03-18
Fix fee calc bug caused by 0 bal transactions in unspent list

carybakker:
### In this PR
 * Defrag integration test was unexpectedly failing with insufficient funds
 * Funds were available - so test should not have failed
 * Inspection revealed the 0-bal transactions inflating the defrag fee cost
 * Proposed fix here is to filter out 0-bal transactions from unspent txos
 * Alternate fix could be to leave the 0-bal transactions  in the list, and update fee calc logic to exclude

fixes #184740756


carybakker:
Purpose of PR was to enable discussion around the topic - will revisit this shortly with a final actual solution after further investigation

PR #244, opened 2023-03-22
Switch gRPCs cert-pinning to shared implementation

the-real-adammork:
Soundtrack of this PR: [Normon Connors - She's Gone](https://www.youtube.com/watch?v=LPcYX0G2UsE)

### Motivation

gRPC's built-in cert-pinning doesn't like the alpha network. We've debugged and we see the correct certs from the server so its unclear why the built-in verification fails. HTTP cert-pinning code works for alpha net.

### In this PR
* Factor out HTTP cert-pinning code into shared functions
* Use shared cert-pinning code in gRPC custom verification callback 
* Add unit tests for shared cert-pinning code.

### Future Work
* Maybe figure out why gRPC doesn't like the alpha network ?


carybakker (approved):
LGTM

PR #249, opened 2023-04-27
Cary/dynamic test wallets 184740454

carybakker:

### Motivation

Integration tests require the use of test wallets and on-chain transactions.  Tests running in parallel require unique funded wallet addresses to avoid spend collisions (i.e. "input already spent" errors).  As we have a finite number of funded accounts available, this makes running the integration tests difficult when we have both developers and CICD potentially running integration tests at the same time.  The solution is to create test wallets on the fly as needed by the tests, where each developer or environment has their own source account to fund the dynamically created accounts - thereby avoiding usage collisions.

### In this PR
* Pivotal Tracker ID 184740454
* A separate test app under ./tools/ that is capable of funding a series of wallets from a configured source funding account, all from a single seed value, and all recoverable from the single seed value.
* updated github workflow to run the dynamic test account creation job prior to running the integration tests
* updated integration tests logic to be able to look up the specifically created wallets required based on the specific test name
* removed the forced delay between protocol test that was previously required for back-to-back protocol tests to not 'collide' by using same inputs.  Now the tests can run immediately without delay.

### Future Work
 * Improve configuration of required test wallets so as to not have the required configuration in the tools/ app (possibly extract the data from comments embedded in the tests themselves?
 * Explore re-combining the test account creation with the Integration Tests
 * Ability to fund a single, individual test, on the fly immediately prior to test execution - for development iteration
 * Re-writing the defrag test


Tests/Integration/Transacting/MobileCoinClientPublicApiIntTests.swift
```diff
@@ -14,33 +14,40 @@ import XCTest
...
-        let client = try await IntegrationTestFixtures.createMobileCoinClientWithBalance(
-            accountKey: accountKey,
-            transportProtocol: transportProtocol
-        )
-        let transaction = try await client.prepareTransaction(
-            to: recipient,
-            amount: Amount(100, in: .MOB),
-            fee: IntegrationTestFixtures.fee)
-        try await client.submitTransaction(transaction: transaction.transaction)
+//        let (_, client) = try IntegrationTestFixtures.createDynamicClient(
```
the-real-adammork: is this expected ?

the-real-adammork:
do you know why CD is failing ?

the-real-adammork (approved):

PR #278, opened 2024-11-05
chore: update for SPM w/ Xcode 16 support

the-real-adammork:
### Motivation

Need SPM Support

### In this PR
*  Updates to support the latest Xcode and SPM

### Future Work
* Figure out fix for git-lfs in SPM - https://forums.swift.org/t/swiftpm-with-git-lfs/42396/29


the-real-adammork:
Fix CI in follow-up

PR #279, opened 2025-03-13
 Add API for creating a Proof of Reserve SCI

eranrund:
### Motivation

We want Sentz to be able to generate Proof of Reserve SCIs. In order to do that, the `mobilecoin_flutter_plugin` needs to support that, and in order for it to support it, the underlying `MobileCoin-Swift` (and [`android-sdk`](https://github.com/mobilecoinofficial/android-sdk/pull/144)) components need to support that.

### In this PR
* A helper method for creating a Proof of Reserve SCI
* I had to rename a bunch of `serializedBytes: ` instances to `serializedData: ` since the code wouldn't build without this.

### DIFF
--- a/ExampleHTTP/Podfile.lock
+++ b/ExampleHTTP/Podfile.lock
@@ -1,6 +1,6 @@
 PODS:
   - Keys (1.0.1)
-  - LibMobileCoin/CoreHTTP (6.0.0-pre1):
+  - LibMobileCoin/CoreHTTP (6.0.2):
     - SwiftProtobuf (~> 1.5)
   - Logging (1.4.0)
   - MobileCoin (6.0.3)
@@ -48,7 +48,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   Keys: a576f4c9c1c641ca913a959a9c62ed3f215a8de9
-  LibMobileCoin: 4b89263fc8280242da38ecaf8cbcfa473822c260
+  LibMobileCoin: 8503f567fa32184a5be7bc038fbd727747dd9991
   Logging: beeb016c9c80cf77042d62e83495816847ef108b
   MobileCoin: c64e04cf7eaa17d83efb2f8d69c02d1d53889ffa
   SwiftLint: f80f1be7fa96d30e0aa68e58d45d4ea1ccaac519
--- a/Sources/Common/Transaction/SignedContingentInput.swift
+++ b/Sources/Common/Transaction/SignedContingentInput.swift
@@ -19,7 +19,7 @@ public struct SignedContingentInput {
 
     /// - Returns: `nil` when the input is not deserializable.
     public init?(serializedData: Data) {
-        guard let proto = try? External_SignedContingentInput(serializedBytes: serializedData) else {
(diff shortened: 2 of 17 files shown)


ExampleHTTP/Podfile.lock
```diff
@@ -1,6 +1,6 @@
 PODS:
   - Keys (1.0.1)
-  - LibMobileCoin/CoreHTTP (6.0.0-pre1):
+  - LibMobileCoin/CoreHTTP (6.0.2):
```
the-real-adammork, line 3: this needs a PR and release to cocoapods, or we can use its commit hash

Sources/Common/Transaction/SignedContingentInput.swift
```diff
@@ -86,7 +86,15 @@ extension SignedContingentInput {
         self.changeAmount = requiredOutputAmounts.first(where: { $0.tokenId == changeTokenId }) ??
             Amount(0, in: changeTokenId )
 
-        self.rewardAmount = Amount(pseudoOutputAmount.value - changeAmount.value, in: changeTokenId)
+        // Generally the amount we are getting from the SCI (pseudoOutputAmount) will be greater than the change
+        // amount, otherwise there is no point in using it.
+        // However, there is one special case where that is not the case: Proof of Reserve SCIs.
+        // For proof of reserve SCIs the change amount ends up being set to u64::MAX so make them unspendable,
```
the-real-adammork:
```suggestion
        // For proof of reserve SCIs the change amount ends up being set to u64::MAX to make them unspendable,
```

ExampleHTTP/Podfile.lock
eranrund: Hm, I don't think I manually changed this...

the-real-adammork:
Looks good, prolly easiest for me to make a release from my local machine than fix the CI, which should be moved to github-workflows anyways.

the-real-adammork (approved):

ExampleHTTP/Podfile.lock
eranrund: It looks like this has been around for awhile - https://github.com/mobilecoinofficial/libmobilecoin/blob/master/LibMobileCoin.podspec (last changed 10 months ago).
I'm guessing it did get released?

the-real-adammork: maybe it didn't get checked in and the submodule commit hash points to that version, so the lock file got updated.

the-real-adammork: ya that must be the case. so its ok to leave as-is 6.0.2 is already released

eranrund:
> Looks good, prolly easiest for me to make a release from my local machine than fix the CI, which should be moved to github-workflows anyways.

Thanks.
I'm okay with that if you feel comfortable with that. Have you released this in the past since CI became out-of-date/dysfunctional?

the-real-adammork:
> > Looks good, prolly easiest for me to make a release from my local machine than fix the CI, which should be moved to github-workflows anyways.
> 
> Thanks. I'm okay with that if you feel comfortable with that. Have you released this in the past since CI became out-of-date/dysfunctional?

ya I made the last release locally.

PR #284, opened 2026-08-18
chore(SENTZ-5436): bump pod version to 6.0.6

bcgreijnlautier:
* Bump s.version in MobileCoin.podspec to 6.0.6.
* Regenerate ExampleHTTP/Podfile.lock in lockstep, since CI installs pods in deployment mode and fails on a stale lock.


stevenspiel (approved):
LGTM

PR #286, opened 2026-08-26
Derive pub key from rng seed

stevenspiel:
Currently, the only way to get a txout's public key is by calling prepareTransaction, which selects inputs and consequently needs a funded, synced account. But it shouldn't need that.

Inputs are added at [`TransactionBuilder.swift:211`](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/4f261da84c3e3ae1dc6e20b8df42d57aec4c42de/Sources/Common/Transaction/TransactionBuilder.swift#L211) — *before* the seeded RNG is even created at [`:216`](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/4f261da84c3e3ae1dc6e20b8df42d57aec4c42de/Sources/Common/Transaction/TransactionBuilder.swift#L216) — and consume none of it. So the keys a seed produces have never depended on what is being spent.

## What this adds

`TransactionBuilder.txOutContexts` runs the output-adding against a builder with no inputs and never calls `build()`. `MobileCoinClient.txOutContexts` wraps it: same `generateRngSeed()` hop as `prepareTransaction`, fog resolved for the recipient and the account's own address, block version fetched, then delegate.

## Why it's an extraction, not a parallel path

The point is that the two paths *cannot* disagree, so both shared dependencies are now extracted rather than duplicated:

| Extracted | Why it matters to the keys |
| --- | --- |
| `addOutputs` | Holds the payload loop and the existing [`changeContext`](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/4f261da84c3e3ae1dc6e20b8df42d57aec4c42de/Sources/Common/Transaction/TransactionBuilder.swift#L270) branch, so every draw before `build()` happens in one place |
| `makeBuilder` | `build` and the new path had been constructing the builder identically by hand. The fog resolver decides real hint vs fake hint — different numbers of draws — and the block version decides where change goes |

That is not theoretical. In core, [`create_fog_hint` draws before the tx private key](https://github.com/mobilecoinfoundation/mobilecoin/blob/fe8017c7067b6c4c116f353b9e1beb3db5639888/transaction/builder/src/transaction_builder.rs#L670), and [`r` is drawn immediately after](https://github.com/mobilecoinfoundation/mobilecoin/blob/fe8017c7067b6c4c116f353b9e1beb3db5639888/transaction/builder/src/transaction_builder.rs#L681), so a resolver built differently shifts every key that follows.

For the same reason `MobileCoinClient.txOutContexts` resolves fog through the `desiredMinPubkeyExpiry` overload, matching [`TransactionPreparer.swift:137`](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/4f261da84c3e3ae1dc6e20b8df42d57aec4c42de/Sources/Common/Transaction/TransactionPreparer.swift#L137) rather than the plainer one. An earlier revision used the plain overload and would have returned keys the built transaction did not carry.

## The zero change amount

The derivation reports zero change, having no inputs to leave a remainder, and still has to match a real transaction's keys.

`testChangeAmountDoesNotMoveTheChangeKey` proves it: the built transaction's change is asserted non-zero so the test cannot pass vacuously, and the change keys match anyway. **Amounts never reach the RNG.** That was the load-bearing assumption behind the zero, and it is now demonstrated rather than argued.

## Cross-platform parity

Both SDKs derive the builder seed from the caller's RNG differently — [`generateRngSeed()`](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/4f261da84c3e3ae1dc6e20b8df42d57aec4c42de/Sources/Common/Utils/Rng/MobileCoinRng.swift#L15) is four `next()` UInt64s reduced into `Data`, where android-sdk takes `nextBytes(32)`. They should agree, since `next_u64` combines two u32 words as `w[i+1] << 32 | w[i]` and its little-endian bytes are what `fill_bytes` writes for those words, but nothing had checked: idempotent retries always happen on the device that created the seed.

That hop is the only platform-specific code in the chain, so `SeedableRngUnitTests.testBuilderSeedIsPlatformIndependent` pins it with a golden value. android-sdk asserts the same constant against the same seed. **The two must change together or not at all.**

### DIFF
--- a/Sources/Common/MobileCoinClient.swift
+++ b/Sources/Common/MobileCoinClient.swift
@@ -395,6 +395,90 @@ public final class MobileCoinClient {
         }
     }
 
+    /// Derives the payload and change TxOut public keys a transaction built
+    /// with `rng` would produce, without building one.
+    ///
+    /// Intended for learning a TxOut public key before the transaction exists
+    /// — for instance to tell a third party which output to expect. Passing
+    /// the same `rng` to `prepareTransaction` later produces the same keys.
+    ///
+    /// No balance is required: inputs are what need funds, and none are added.
+    /// Fog reports and the block version are fetched, so this makes network
+    /// calls and can fail like any other fog operation.
+    ///
+    /// `amount` and `fee` are what the eventual transaction will carry. They
+    /// do not affect the keys, but the recipient does, and so does `memoType`
+    /// insofar as it must match what the transaction is built with.
+    public func txOutContexts(
+        to recipient: PublicAddress,
+        memoType: MemoType = .recoverable,
+        amount: Amount,
+        fee: UInt64,
+        rng: MobileCoinRng,
+        completion: @escaping (
+            Result<(payload: TxOutContext, change: TxOutContext), TransactionPreparationError>
+        ) -> Void
+    ) {
+        guard let rngSeed = rng.generateRngSeed() else {
+            completion(.failure(
+                TransactionPreparationError.invalidInput("Could not create 32-byte RNG seed")))
+            return
+        }
+
+        let (accountKey, ledgerBlockCount) = accountLock.readSync {
+            ($0.accountKey, $0.knowableBlockCount)
+        }
+
+        // Same overload and arguments the real path uses in
+        // TransactionPreparer: the expiry decides which fog pubkeys resolve,
+        // and that decides whether each output draws a real hint or a fake
+        // one, which are different numbers of draws.
+        let tombstoneBlockIndex = ledgerBlockCount + 50
+
+        fogResolverManager.fogResolver(
+            addresses: [recipient, accountKey.publicAddress],
+            desiredMinPubkeyExpiry: tombstoneBlockIndex
+        ) { fogResolverResult in
+            switch fogResolverResult {
+            case .failure(let error):
+                self.callbackQueue.async {
+                    completion(.failure(.connectionError(error)))
+                }
+            case .success(let fogResolver):
+                self.metaFetcher.blockVersion { blockVersionResult in
+                    switch blockVersionResult {
+                    case .failure(let error):
+                        self.callbackQueue.async {
+                            completion(.failure(.connectionError(error)))
+                        }
+                    case .success(let blockVersion):
+                        let context = TransactionBuilder.Context(
+                            accountKey: accountKey,
+                            blockVersion: blockVersion,
+                            fogResolver: fogResolver,
+                            memoType: memoType,
+                            tombstoneBlockIndex: tombstoneBlockIndex,
+                            fee: Amount(fee, in: amount.tokenId),
+                            rngSeed: rngSeed)
+
+                        let result = TransactionBuilder.txOutContexts(
+                            context: context,
+                            recipient: recipient,
+                            amount: amount
+                        ).mapError { error in
+                            TransactionPreparationError.invalidInput(
+                                error.localizedDescription)
+                        }
+
+                        self.callbackQueue.async {
+                            completion(result)
+                        }
+                    }
+                }
+            }
+        }
+    }
+
     public func prepareDefragmentationStepTransactions(
         toSendAmount amount: Amount,
         recoverableMemo: Bool = false,
--- a/Sources/Common/MobileCoinClient.swift
+++ b/Sources/Common/MobileCoinClient.swift
@@ -444,7 +444,6 @@ public final class MobileCoinClient {
             case .success(let fogResolver):
                 self.deriveTxOutContexts(
                     to: recipient,
-                    accountKey: accountKey,
                     fogResolver: fogResolver,
                     tombstoneBlockIndex: tombstoneBlockIndex,
                     rngSeed: rngSeed,
@@ -455,7 +454,6 @@ public final class MobileCoinClient {
 
     private func deriveTxOutContexts(
         to recipient: PublicAddress,
-        accountKey: AccountKey,
         fogResolver: FogResolver,
         tombstoneBlockIndex: UInt64,
         rngSeed: RngSeed,
@@ -469,7 +467,7 @@ public final class MobileCoinClient {
                 self.finish(.failure(.connectionError(error)), completion)
             case .success(let blockVersion):
                 let context = TransactionBuilder.Context(
-                    accountKey: accountKey,
+                    accountKey: self.accountLock.readSync { $0.accountKey },
                     blockVersion: blockVersion,
                     fogResolver: fogResolver,
                     memoType: .recoverable,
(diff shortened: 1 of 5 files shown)


Tests/Unit/Transaction/TxOutContextsDerivationTests.swift
```diff
@@ -0,0 +1,131 @@
...
+        let output = try XCTUnwrap(fixture.outputs.first)
+
+        let first = try XCTUnwrapSuccess(TransactionBuilder.txOutContexts(
+            context: Self.context(fixture: fixture, rngSeed: RngSeed()),
+            recipient: output.recipient,
+            amount: output.amount))
+        let second = try XCTUnwrapSuccess(TransactionBuilder.txOutContexts(
+            context: Self.context(fixture: fixture, rngSeed: RngSeed()),
+            recipient: output.recipient,
+            amount: output.amount))
```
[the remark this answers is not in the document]
stevenspiel: 810871f4dabffa731340d25b48bb405281cd26f1

Sources/Common/Transaction/TransactionBuilder.swift
```diff
@@ -267,6 +250,110 @@ extension TransactionBuilder {
...
+        case .failure(let error):
+            return .failure(error)
+        case .success(let made):
+            builder = made
+        }
+
+        let possibleTransaction = PossibleTransaction(
+            [TransactionOutput(recipient: recipient, amount: amount)],
+            Amount(0, in: context.fee.tokenId))
+
```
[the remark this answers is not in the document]
stevenspiel: 9c83214ca5353c95492f08c7029536eeb84afb27

Sources/Common/MobileCoinClient.swift
```diff
@@ -395,6 +395,90 @@ public final class MobileCoinClient {
         }
     }
 
+    /// Derives the payload and change TxOut public keys a transaction built
+    /// with `rng` would produce, without building one.
+    ///
+    /// Intended for learning a TxOut public key before the transaction exists
+    /// — for instance to tell a third party which output to expect. Passing
+    /// the same `rng` to `prepareTransaction` later produces the same keys.
```
bcgreijnlautier: potential issue: `generateRngSeed()` consumes four `next()` draws, so handing the same `MobileCoinRng` instance to `prepareTransaction` afterwards derives a different seed and different keys than the ones returned here. A caller who follows this line would tell a third party a TxOut public key the transaction doesn't contain. I think the contract has to be a fresh rng rebuilt from the same `RngSeed`, and that should probably be specified here.

Sources/Common/MobileCoinClient.swift
```diff
@@ -395,6 +395,90 @@ public final class MobileCoinClient {
...
+
+        let (accountKey, ledgerBlockCount) = accountLock.readSync {
+            ($0.accountKey, $0.knowableBlockCount)
+        }
+
+        // Same overload and arguments the real path uses in
+        // TransactionPreparer: the expiry decides which fog pubkeys resolve,
+        // and that decides whether each output draws a real hint or a fake
+        // one, which are different numbers of draws.
+        let tombstoneBlockIndex = ledgerBlockCount + 50
```
bcgreijnlautier, line 455: question: this expiry comes from the ledger count at derive time, while `prepareTransaction` recomputes it from a larger count when the send actually happens, so the two calls do not pass the same `desiredMinPubkeyExpiry` after all. By the reasoning in the comment above, a different fog pubkey resolving between them could flip a hint and move every draw after it. `knowableBlockCount` is also 0 before the client syncs, which the "no balance is required" doc reads as an invitation to hit. Is the draw count actually stable across all that, or does the caller need to derive and send close together?

Sources/Common/MobileCoinClient.swift
```diff
@@ -395,6 +395,90 @@ public final class MobileCoinClient {
...
+                    case .success(let blockVersion):
+                        let context = TransactionBuilder.Context(
+                            accountKey: accountKey,
+                            blockVersion: blockVersion,
+                            fogResolver: fogResolver,
+                            memoType: .recoverable,
+                            tombstoneBlockIndex: tombstoneBlockIndex,
+                            // No amount reaches a draw, so any fee gives the
+                            // same keys; zero avoids implying otherwise.
+                            fee: Amount(0, in: .MOB),
```
bcgreijnlautier: suggestion: this pins every derivation to `.MOB`, and `TransactionBuilder.txOutContexts` takes both zero amounts from `context.fee.tokenId`, so deriving for a non-MOB transaction runs the builder with a token id the real send will not use. `testFeeDoesNotMoveTheKeys` varies only the value and keeps `fixture.fee.tokenId`, so nothing pins that the token id is as irrelevant to the keys as the value is. I would either take the token id as a parameter here or extend that test to vary it.

Sources/Common/MobileCoinClient.swift
```diff
@@ -395,6 +395,90 @@ public final class MobileCoinClient {
...
+                            rngSeed: rngSeed)
+
+                        let result = TransactionBuilder.txOutContexts(
+                            context: context,
+                            recipient: recipient
+                        ).mapError { error in
+                            TransactionPreparationError.invalidInput(
+                                error.localizedDescription)
```
bcgreijnlautier: suggestion: `TransactionBuilderError` is not a `LocalizedError`, so `localizedDescription` drops the reason and returns the Foundation default string.

```suggestion
                                String(describing: error))
```

bcgreijnlautier (approved):

Sources/Common/MobileCoinClient.swift
stevenspiel: yep. great catch! 5b63a9d600fb31b9eebf24edb97ed4721d8ffb16 I changed the arg to be an RngSeed, rather than an rng

stevenspiel: changing this on the android side, too, so the APIs match

Sources/Common/MobileCoinClient.swift
stevenspiel: The expiries don't need to match. I've updated the comment to reflect. That comment was based on a previous finding which I proved false. 23bfa89a95d42f3ff120fa1c9a31b87dbb717e6d

Sources/Common/MobileCoinClient.swift
stevenspiel: added a test to cover that case ff9f54b3118a45e2bcee6c53ee7c371aa1760fde

Sources/Common/MobileCoinClient.swift
stevenspiel: 15ce5d5f6a0593232648635c99bf16322c99cf62

Sources/Common/MobileCoinClient.swift
stevenspiel: and a case to cover that the block version does not move the keys, either ad774262f3e95361d498393b6d170c642ec67021

PR #287, opened 2026-08-29
chore(SENTZ-5520): default the SwiftPM route to the HTTP transport

bcgreijnlautier:
Soundtrack of this PR: [Talking Heads - Once in a Lifetime](https://www.youtube.com/results?search_query=talking+heads+once+in+a+lifetime)

### Motivation

Currently the SwiftPM route defaults to the gRPC transport, and the live app path is HTTP only. A SwiftPM consumer compiles and links the whole gRPC and NIO stack for a transport it never uses.

### In this PR
* Make the SwiftPM `MobileCoinCore` product HTTP only. It takes `LibMobileCoinCoreHTTP` and its file list mirrors the podspec `CoreHTTP` subspec, so both routes ship the same code.
* Leave `Sources/GRPC` out of the target rather than guarding it, because its files import GRPC and NIO unconditionally. `TransportProtocol.grpc` now returns a `connectionFailure` instead of never calling back, and `supportedProtocols` returns `[.http]`.
* Exclude `Tests/ProtocolSpecific/Grpc` and build the test targets in `build-spm`. The test target stopped compiling when the dependency closure lost NIOHPACK, and plain `swift build` cannot see that.
* Delete `GRPCOnly`. No build route compiles it, and its only reference in the repository was the SwiftLint include added by this branch. Two of its four files duplicate declarations `Sources/Common` now carries, and `WrappedSSLCertificateValidator` returns a type that exists nowhere else, so it could never have compiled.
* Delete the three `CocoapodsOnly/Common/Utils` bridging-header stand-ins and the `CocoapodsOnly/*.{h,m,swift}` glob that listed them. The repository declares no `SWIFT_OBJC_BRIDGING_HEADER` anywhere, nothing includes `MobileCoin-Bridging-Header.h`, and the only includes of `TestBridgingHeader.h` come from two siblings in the same directory. The subspec now lists only its Swift sources.
* Update the `MobileCoin` entry in `ExampleHTTP/Podfile.lock`'s `SPEC CHECKSUMS`. The lock records the SHA1 of the podspec JSON CocoaPods generates, so a podspec edit without it fails CI's `pod install --deployment` on a whole-lockfile comparison, reported as a lockfile delta with nothing in the podspec to point at.
* Point the last `Example/` references at `ExampleHTTP`, in `.jazzy.yaml`, the README directory table, and `scripts/reinitialize_keys`. That script has been dead since `db7782e`: it runs under `set -e` and both secrets helpers `pushd` into their argument, so it exited on the missing `Example/` before ever reaching the ExampleHTTP pair.
* Drop the dependabot bundler entry for `/Example` rather than moving it to `/ExampleHTTP`. That Gemfile holds `json` at 2.6.3 because the value decides the `Podfile.lock` checksums above, and dependabot updates an exact pin like any other, so aiming a stream at it means ignoring `json` by name. The entry has produced nothing for two years and the root entry already covers the gems both Gemfiles share.

Resolution still fetches grpc-swift and NIO, because libmobilecoin 6.1.0 declares grpc-swift in its own manifest. What goes is the compile and the link.

Three `MistySwapTests` hard-code `.grpc`. They sit behind CI-only secrets, so the local gate never runs them and the first CI run after merge is what proves them either way.

Implements SENTZ-5520.

### Future Work
* SENTZ-5530 - `tools/TestSetupClient` still requests `.grpc` and does not build against `master` in any case.


.swiftlint.yml
```diff
@@ -95,6 +95,7 @@ opt_in_rules: # some rules are only opt-in
 included: # paths to include during linting. Ignored if `--path` is present unless `--force-exclude` is specified.
   - Sources
   - Tests
+  - GRPCOnly
```
stevenspiel: /GrpcOnly seems to now be dead code (only referenced here). Can we just remove it?

CHANGELOG.md
```diff
@@ -5,6 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
stevenspiel: should this have a version number?

```suggestion
## [6.1.1] - 2026-09-02
```

stevenspiel (approved):
LGTM

CHANGELOG.md
bcgreijnlautier: Yep, folded into the existing `## [6.1.0]` section instead

PR #289, opened 2026-08-29
chore(SENTZ-5521): adopt the Swift 6 language mode on the library targets

bcgreijnlautier:
Soundtrack of this PR: [New Order - Blue Monday](https://www.youtube.com/results?search_query=new+order+blue+monday)

### Motivation

Currently this package compiles in the Swift 5 language mode, so the data races the compiler can now prove stay invisible. Moving the library to Swift 6 makes them visible, and it surfaced two real ones.

### In this PR
* Raise swift-tools to 6.1, so the library takes the Swift 6 language mode. The two test targets hold at v5 by an explicit setting, which SENTZ-5527 removes.
* Annotate the SDK's `Sendable` and immutable global state across `Sources/Common`. The two lock types take `@unchecked Sendable where Value: Sendable`, since the queue guards the value and the compiler cannot see that.
* Move certificate pinning out of `DefaultHttpRequester` into a `CertificatePinningDelegate` holding both trust roots behind one `ReadWriteDispatchLock`. URLSession calls the delegate on its own queue while the SDK can be setting roots from another thread. `DefaultHttpRequester` becomes `final` and stops conforming to `URLSessionDelegate` itself, which is source breaking for a subclass. It invalidates its session on `deinit`, because a URLSession holds its delegate until it is invalidated, and a new test asserts the delegate is released once the requester goes out of scope.
* Drop the `lazy var` cache in `ImmutableOnceReadLock.get()`. Lazy initialization carries no synchronization of its own, so a concurrent first read is a race. Every read goes through the lock now.
* Import Security with `@preconcurrency`, so `SecSSLCertificates` can be `Sendable`. `SecKey` is a CoreFoundation type and carries no conformance of its own.
* Bound SwiftProtobuf at `>= 1.36, < 1.38` in the podspec and raise the SwiftPM floor to 1.36.1. 1.38 emits `nonisolated extension`, which pre-Swift-6.1 toolchains reject. The `MobileCoin` entry in `ExampleHTTP/Podfile.lock` moves with it, since that checksum is the SHA1 of the generated podspec JSON.

The two routes now have different toolchain floors, and only SwiftPM's manifest enforces its own: a `swift-tools-version:6.1` manifest is rejected outright by anything older. The CocoaPods floor of Swift 5.10 is set by `nonisolated(unsafe)` in `MobileCoinLogging`, which no earlier compiler parses, and the podspec's `swift_version` is a language mode that expresses none of it. The `#error` guard added below that floor reports the requirement rather than imposing it, in place of a bare syntax error naming neither the cause nor the version. The CHANGELOG records both floors and every source break above, and the README now carries a Toolchain table naming both. The setup steps move to Ruby 3.1.x, which is what activesupport 7.2.2.1 in the committed lockfiles requires.

Implements SENTZ-5521.

### Future Work
* SENTZ-5527 - move the two test targets off the v5 language mode.


CHANGELOG.md
```diff
@@ -12,6 +12,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The SwiftPM `MobileCoinCore` product is now HTTP only. It consumes
   `LibMobileCoinCoreHTTP` and no longer compiles `Sources/GRPC`. This matches
   the CocoaPods chain, whose only subspec has been `CoreHTTP` for some time.
+- The library compiles in the Swift 6 language mode. The two test targets hold
+  at Swift 5.
+- The two routes now have different toolchain floors. SwiftPM needs Swift 6.1,
+  Xcode 16.3 or newer, because the manifest declares
```
stevenspiel, line 18: flagging that  CI runs `macos-15` (Xcode 16) and will need to be upgraded. 

Also, maybe this now belongs in the README and possibly in the code itself. I'll find those places and add suggestions

stevenspiel: readme is not included as a changed file here, but something like 

```
## Build Instructions

### Toolchain

The two routes have different floors, and neither manifest can express them:

| route | minimum | why |
|---|---|---|
| SwiftPM | Swift 6.1 / Xcode 16.3 | `Package.swift` declares `swift-tools-version:6.1`; older SwiftPM rejects the manifest before reading anything else |
| CocoaPods | Swift 5.10 / Xcode 15.3 | `MobileCoinLogging` uses `nonisolated(unsafe)` |

The podspec's `swift_version` is a language mode, not a compiler floor, so it
enforces neither.

The workspace can be built with `make`.
```

Sources/Common/Shared/MobileCoinLogging.swift
```diff
@@ -7,7 +7,11 @@
 import Foundation
```
stevenspiel, line 16:
```suggestion
//
//  Copyright (c) 2020-2026 MobileCoin. All rights reserved.
//

// The `nonisolated(unsafe)` below is what sets the CocoaPods floor: no earlier
// compiler parses it, and the resulting syntax error names neither the cause
// nor the requirement. SwiftPM consumers never reach this file — the 6.1
// manifest is rejected first — so this diagnostic is for the pod route.
#if compiler(<5.10)
#error("MobileCoin requires Swift 5.10 / Xcode 15.3 via CocoaPods, or Swift 6.1 / Xcode 16.3 via SwiftPM.")
#endif

// swiftlint:disable prefixed_toplevel_constant

import Foundation
```

stevenspiel (approved):
LGTM

CHANGELOG.md
bcgreijnlautier: Added. The CI is already running on macos-26 / Xcode 26.6, just needed a doc update

PR #290, opened 2026-08-29
ci(SENTZ-5517): raise the pod iOS deployment floor to 13

bcgreijnlautier:
Soundtrack of this PR: [Bob Dylan - The Times They Are a-Changin'](https://www.youtube.com/results?search_query=bob+dylan+the+times+they+are+a+changin)

### Motivation

libmobilecoin raises its pod iOS deployment floor to 13.0 under SENTZ-5517. CocoaPods refuses a dependency whose deployment target is above the dependent's, so this pod moves with it or the pair stops installing. This PR is independent of the SENTZ-5520 stack in this repository.

### In this PR
* Raise the pod iOS deployment floor to 13.0, which drops iOS 12 consumers.
* Bump `s.version` to 6.1.0. It was 6.0.6, and v6.0.6 is already tagged at `4f261da`. `pod trunk push` would have published a spec whose `:tag => "v#{s.version}"` points at a tree without the new floor. A minor rather than a patch, because raising a deployment target would otherwise drag every downstream `~> 6.0.x` pinner onto a floor it never asked for.
* Point the LibMobileCoin dependency at `>= 6.1.0`, `< 7.0`. It was `~> 6.0.0-pre1`, which caps below 6.1.0 and so can never pick up the release carrying the matching floor.

This installs once libmobilecoin's stack lands and cuts v6.1.0.

Implements SENTZ-5517.

### Future Work
* SENTZ-5533 - the `Example` projects still target 12.2, and nothing in CI asserts the two podspec floors agree.
* SENTZ-5532 - `mobilecoin_flutter_plugin` pins SwiftProtobuf 1.29.0, which 6.1.0 makes unsatisfiable.


bcgreijnlautier:
Superseded by #293, which raises the libmobilecoin floor to 6.1.0 and carries the deployment-target change with it. SENTZ-5564 records this work as superseded.

bcgreijnlautier:
Correction to my closing comment above: #293 does not carry the deployment-target change. Its diff touches ci.yml, .gitmodules, .swiftlint.yml, Makefile, Package.resolved, Package.swift, README.md and Base58CoderTests.swift, and a grep for `deployment_target` or `platform :ios` across that range returns nothing. `MobileCoin.podspec:23` still reads `s.platform = :ios, "12.2"` at #293's tip.

This work is a prerequisite for the CocoaPods side of the 6.1.0 pin, not superseded by it. LibMobileCoin 6.1.0 declares an iOS 13.0 floor, and `MobileCoin.podspec` still caps its dependency at `~> 6.0.0-pre1`, which excludes 6.1.0. Both commits now ride on SENTZ-5560/chore/drop-libmobilecoin-submodule, which is where the LibMobileCoin 6.1.0 pod pin lands. This PR stays closed because the commits ship there, not because the change was superseded.

PR #292, opened 2026-08-31
chore: bump pod version to 6.0.7

stevenspiel:
## What this releases

#286: `MobileCoinClient.txOutContexts(to:rngSeed:)`, which derives the payload and change TxOut public keys a transaction will carry from an RNG seed alone, without building or submitting one. The MistySign offramp needs a `mobilecoin_txo_pub_key` before the transaction exists, and the plugin's `getTxOutPublicKeys` bridge is waiting on this pod.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.0.7] - 2026-08-31
+
+### Added
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.0.7] - 2026-08-31
```
bcgreijnlautier: nit: this lands directly on top of `## [4.0.1] - 2023-03-02`, so every 5.x and 6.0.x release is still missing from the file, including the `MistysignAttestedSession` that 6.0.6 added. Worth backfilling in another PR if we are treating the changelog as current again.

bcgreijnlautier (approved):

PR #293, opened 2026-09-01
chore(SENTZ-5553): Raise the libmobilecoin floor to 6.1.0

bcgreijnlautier:
libmobilecoin v6.0.4 carries two submodules, `Vendor/mobilecoin` for the foundation monorepo and `Artifacts` for a git-lfs repo of prebuilt `.a` files, and every SwiftPM consumer pays both at resolve time. v6.1.0 has no `.gitmodules`: it ships the xcframework as a checksummed release asset through a `binaryTarget`, so the resolve takes the asset and no submodule. The lock also moves swift-protobuf to 1.37.0, because the libmobilecoin change forces a full re-resolve and 6.1.0 caps swift-protobuf below 1.38.0.

`libmobilecoin` then moved. A fresh repository took over the name and the old one became `libmobilecoin-archive`, which killed GitHub's rename redirect and every pin that resolved a pre-6.1.0 object. This branch repins SwiftPM at the new repository's v6.1.0 revision and repoints the `Vendor/libmobilecoin` submodule at the archive that still holds its gitlink.

Notes:

- `pod-lib-lint` and `publish-pod-release` are removed here rather than deferred. `MobileCoin.podspec` pins `LibMobileCoin/CoreHTTP` at `~> 6.0.0-pre1`, CocoaPods takes the highest match, and that is 6.0.2, whose source is the repository name the fresh libmobilecoin now holds. `pod lib lint` reads CocoaPods trunk and nothing else, so the clone fails. Three of the 47 trunk versions carry that source; the rest point at libmobilecoin-ios-artifacts and still resolve. SENTZ-5564 retires the pod rather than republishing it.
- `MobileCoin.podspec` stays. ExampleHTTP consumes it through a local `path:` external source that never reads trunk, and the SwiftLint binary comes from the Pods tree the same `pod install` populates, so deleting it would take out the two jobs that are currently green.
- A new `build-spm` job compiles the package and its test targets without the CocoaPods toolchain, which is what proves an SPM consumer can resolve libmobilecoin on a cold machine. The tests compile but do not run, because generating `Tests/Common/Secrets/secrets.json` needs a macOS-keychain identity.
- mobilecoin-app reaches this repo through CocoaPods today, so the floor does not change the app. SENTZ-5555 moves the Flutter plugin onto SwiftPM.
- `Package.swift` joins the SwiftLint `excluded:` list. SwiftPM fixes the manifest's shape, so `file_header` and `prefixed_toplevel_constant` fire on it and can never be satisfied.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,13 +141,14 @@ upgrade-deps:
 generate-local-process-info:
 	tools/generate_process_info_jsons.sh
 
-# Builds every target in Package.swift. Unlike `run-all-tests-spm` this needs no
-# secrets, so it is the one SPM check CI can run today. The test targets declare
-# resources (secrets.json, process_info.json) that are generated rather than
-# committed; SwiftPM only warns about those at build time.
+# Builds every target in Package.swift, test targets included. Unlike
+# `run-all-tests-spm` this needs no secrets, so it is the one SPM check CI can
+# run today. The test targets declare resources (secrets.json,
+# process_info.json) that are generated rather than committed. SwiftPM only
+# warns about those at build time.
 .PHONY: build-spm
 build-spm:
-	swift build
+	swift build --build-tests
 
 .PHONY: fund-test-wallets-spm
 fund-test-wallets-spm:
--- a/Makefile
+++ b/Makefile
@@ -61,44 +61,20 @@ autocorrect:
 	@PATH="./ExampleHTTP/Pods/SwiftLint:$$PATH" swiftlint --fix
 
 .PHONY: lint-all
-lint-all: lint lint-podspec lint-docs
-
-.PHONY: publish
-publish: tag-release publish-podspec
+lint-all: lint lint-docs
 
 # Release
 
+# The podspec is the version source for the tag.
 .PHONY: tag-release
 tag-release:
 	VERSION="$$(bundle exec pod ipc spec MobileCoin.podspec | jq -r '.version')" && \
-		git tag "v$$VERSION" && \
-		git push origin "refs/tags/v$$VERSION"
-
-# MobileCoin pod
-
-# Dynamic linkage, matching what `pod trunk push` validates at release time.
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -61,44 +61,20 @@ autocorrect:
...
-
-.PHONY: publish
-publish: tag-release publish-podspec
+lint-all: lint lint-docs
 
 # Release
 
+# The podspec is the version source for the tag.
 .PHONY: tag-release
 tag-release:
```
stevenspiel, line 70: question: this doesn't seem to be called anywhere. Is it something that will need to be run manually?

Package.swift
```diff
@@ -18,7 +18,7 @@ let package = Package(
         // and from where they can be fetched:
         .package(
             url: "https://github.com/mobilecoinofficial/libmobilecoin.git",
-            from: "6.0.4"
+            from: "6.1.0"
```
stevenspiel, line 21: `MobileCoin.podspec` [still has](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/SENTZ-5553/chore/libmobilecoin-6-1-floor/MobileCoin.podspec#L64) `LibMobileCoin/CoreHTTP ~> 6.0.0-pre1` (though maybe you're getting rid of the podspec).

stevenspiel (approved):
LGTM

Package.swift
stevenspiel: I see this is resolved in the [next PR](https://github.com/mobilecoinofficial/MobileCoin-Swift/pull/294/changes#diff-65ba44658b1714c8777669c7080af5eca4a71bf25f0672e35f3b00b9dd94fd00R67)

Makefile
bcgreijnlautier: Yes, it's run by hand when cutting a release.

PR #294, opened 2026-09-01
chore(SENTZ-5560): Resolve libmobilecoin from its release instead of the submodule

bcgreijnlautier:
Soundtrack of this PR: Talking Heads, "Burning Down the House"

### Motivation

Currently MobileCoin-Swift carries libmobilecoin as a `Vendor/libmobilecoin` submodule, and every macOS CI job pays a recursive submodule checkout plus a git-lfs fetch to populate it. libmobilecoin 6.1.0 publishes its xcframework as a checksummed GitHub release asset, so both consumers can take it from there: SwiftPM through the `binaryTarget` and CocoaPods through the tagged repository. Neither needs a checkout of the submodule any more.

### In this PR

* Drops the `Vendor/libmobilecoin` submodule and its `.gitmodules` entry.
* Resolves LibMobileCoin from the tagged repository, and reads the unit test vectors from `LibMobileCoin/TestVectors` rather than through the submodule path.
* Raises the podspec to 6.1.0 and the iOS deployment floor to 13.0. LibMobileCoin 6.1.0 declares that floor, and CocoaPods refuses a dependency whose deployment target is above the dependent's, so the pair moves together.
* Drops the submodule and git-lfs steps from the macOS CI jobs.
* Ignores `/Vendor/`, so the stale checkout an existing clone still holds does not fail `scripts/check_dirty_git`.

https://mobilecoin-eng.atlassian.net/browse/SENTZ-5560

### Future Work

* [SENTZ-5564](https://mobilecoin-eng.atlassian.net/browse/SENTZ-5564) retires the MobileCoin pod and moves the release path onto the Swift package. The `pod-lib-lint` and `publish-pod-release` jobs are already gone, dropped on the SENTZ-5553 branch below this one; `MobileCoin.podspec` itself is what remains.
* [SENTZ-5562](https://mobilecoin-eng.atlassian.net/browse/SENTZ-5562) moves the flutter plugin and the app onto SwiftPM for libmobilecoin. The plugin pins `MobileCoin/CoreHTTP ~> 6.0.6` at iOS 15.0, so the floor raise does not reach it yet.

[SENTZ-5564]: https://sentz.atlassian.net/browse/SENTZ-5564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.gitignore
```diff
@@ -27,6 +27,9 @@ node_modules/
 # spm build folder
 .build
 
+# stale submodule checkout an existing clone still holds
+/Vendor/
```
stevenspiel, line 31: question: isn't it just libmobilecoin that we want to ignore?

```suggestion
/Vendor/libmobilecoin/
```

stevenspiel (approved):
LGTM

.gitignore
bcgreijnlautier: `Vendor/` has held two submodule paths, `libmobilecoin` and `libmobilecoin-ios-artifacts`. A clone predating the move still carries both as untracked leftovers, so ignoring the directory covers what one path would miss.

PR #296, opened 2026-09-02
chore(SENTZ-5530): Refresh TestSetupClient for the CoreHTTP-only SDK

bcgreijnlautier:
Soundtrack of this PR: [Talking Heads - Road to Nowhere](https://www.youtube.com/results?search_query=talking+heads+road+to+nowhere)

### Motivation

`tools/TestSetupClient` has not resolved since the SDK went CoreHTTP-only. Its Podfile asked for `MobileCoin/Core`, which pulled a trunk `LibMobileCoin`, and neither exists any more.

### In this PR
* Depend on `MobileCoin/CoreHTTP` by path and on `LibMobileCoin/CoreHTTP` at the v6.1.0 tag. The podspec sets `default_subspec = :none` with CoreHTTP as its only subspec, and trunk tops out at 6.0.2.
* Build the client over `.http`, because there is no gRPC subspec left to link against.
* Nest `TestSetupClientTests` with `inherit! :search_paths`. As a peer target it embedded a second copy of every framework into the test bundle.
* Clamp the pod deployment target to 13.0, matching `ExampleHTTP/Podfile`.
* Regenerate the stale `Podfile.lock`, which still pinned 5.x pods.
* Name each test wallet for the test it serves. Eight labels across two blocks still carried method names the integration tests had since renamed. All 56 entries now agree with the `clientNames` list in `IntegrationTestFixtures` that resolves a wallet, which spells the same names without the `test` prefix because it matches on the helper's `#function` and not on the XCTest method.
* Add `tools` to `.swiftlint.yml`'s `included:`, which had never covered it, then fix what surfaced. The largest is the 271-line `createAccounts` body splitting into a computed `testAccountConfigs` and a `makeConfig()` helper.

`pod install` and `xcodebuild build-for-testing` both succeed. No CI job opens this xcodeproj, so `build-spm` and `lint-strict` are what cover it there. Evidence is on the ticket.

Implements SENTZ-5530.

### Future Work
* SENTZ-5527 - move the two test targets off the v5 language mode.


tools/TestSetupClient/TestSetupClient/TestWalletCreator.swift
```diff
@@ -14,280 +14,296 @@ enum TestWalletCreationError: Error, Equatable {
...
+                    [.MOB: [100 + minFee]]),
+            TestAccountFactory.TestAccountConfig(
+                name: "1_HTTP_testTransactionDoubleSubmissionFails_Recipient",
+                txData: [:]),
+            TestAccountFactory.TestAccountConfig(
+                name: "2_GRPC_testTransactionDoubleSubmissionFails_Client",
+                txData:
+                    [.MOB: [100 + minFee]]),
+            TestAccountFactory.TestAccountConfig(
+                name: "3_GRPC_testTransactionDoubleSubmissionFails_Recipient",
```
stevenspiel, line 36: is grpc implementation going away? or is it just paused? These wallets (and plenty more in this file) look like they're going to get funded (not all of them, and, granted, not much), but never used. Deleting the entries isn't the right choice because the tests find wallet by index number (position), not by name. So they'd all need to be updated to point to the new index. Probably not in scope for this PR, but it begs the question of what we want to do with grpc.

tools/TestSetupClient/TestSetupClient/TestWalletCreator.swift
```diff
@@ -14,280 +14,296 @@ enum TestWalletCreationError: Error, Equatable {
...
+                    [.MOB: [100 + minFee]]),
+            TestAccountFactory.TestAccountConfig(
+                name: "5_HTTP_testTransactionStatusFailsWhenInputsAlreadySpent_Recipient",
+                txData: [:]),
+            TestAccountFactory.TestAccountConfig(
+                name: "6_testTransactionStatusFailsWhenInputsAlreadySpent_Client",
+                txData:
+                    [.MOB: [100 + minFee]]),
+            TestAccountFactory.TestAccountConfig(
+                name: "7_testTransactionStatusFailsWhenInputsAlreadySpent_Recipient",
```
stevenspiel: nit: keep with convention:

```suggestion
                name: "6_GRPC_testTransactionStatusFailsWhenInputsAlreadySpent_Client",
                txData:
                    [.MOB: [100 + minFee]]),
            TestAccountFactory.TestAccountConfig(
                name: "7_GRPC_testTransactionStatusFailsWhenInputsAlreadySpent_Recipient",
```

tools/TestSetupClient/TestSetupClientTests/TestSetupClientTests.swift
```diff
@@ -86,7 +83,7 @@ struct ProcessInfoLocal: Decodable {
                 "Or, make duplicate `process_info.json.sample` and remove the `.sample` extension."
             )
         }
-        
+
```
stevenspiel, line 86: thanks for cleaning up the whitespace

stevenspiel (approved):
LGTM

tools/TestSetupClient/TestSetupClient/TestWalletCreator.swift
bcgreijnlautier: Dead in practice. The gRPC subspec was removed as temporary two years ago. `LibMobileCoin` now publishes only `TestVectors` and `CoreHTTP`, so no gRPC artifact is left to link against.

PR #297, opened 2026-09-02
chore(SENTZ-5527): Drop the Swift 5 language mode from the test targets

bcgreijnlautier:
Soundtrack of this PR: [Kraftwerk - The Robots](https://www.youtube.com/results?search_query=kraftwerk+the+robots)

### Motivation

SENTZ-5521 moved the library to the Swift 6 language mode. It held the two test targets at v5, so a data race the compiler can prove in a test goes unseen. This drops that setting. Every target in the package now compiles in one language mode.

### In this PR
* Remove the `swiftLanguageMode(.v5)` setting from both test targets. No target declares a language mode of its own now.
* Make the shared fixtures and mocks concurrency-safe, and give each test class an `@unchecked Sendable` conformance instead of `@MainActor`. A closure written in a test method inherits the class's isolation. `@MainActor` then tells the compiler that lock and queue callbacks run on the main actor. They run on a private queue.
* Take `nonisolated(unsafe)` locals at the two main-queue hops in `SerialCallbackQueueTests`, because `SerialCallbackQueue` declares its continuation without `@Sendable` and runs one task at a time.
* Delete the unreachable `URLSession` delegate code from `MockFailingHttpRequester`, including two typealiases that duplicate the live ones in Sources.

CocoaPods now needs Swift 5.10, Xcode 15.3 or newer, because the shared test mocks use `nonisolated(unsafe)` and no earlier compiler parses it. The CHANGELOG records the floor.

Implements SENTZ-5527.

### Future Work
* SENTZ-5564 - retire the MobileCoin pod and move CI onto the Swift package, which is what removes the CocoaPods compiler floor above.


CHANGELOG.md
```diff
@@ -12,15 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The SwiftPM `MobileCoinCore` product is now HTTP only. It consumes
   `LibMobileCoinCoreHTTP` and no longer compiles `Sources/GRPC`. This matches
   the CocoaPods chain, whose only subspec has been `CoreHTTP` for some time.
-- The library compiles in the Swift 6 language mode. The two test targets hold
-  at Swift 5.
+- Every target in the package compiles in the Swift 6 language mode. No target
+  declares a language mode of its own.
```
stevenspiel: Perhaps more accurately:

```suggestion
- Every target in the package compiles in the Swift 6 language mode. No target
  declares a language mode of its own. The test targets reach it by asserting
  `@unchecked Sendable` on their `XCTestCase` subclasses rather than by
  restructuring how async tests capture `self`. XCTest gives each test method
  its own instance, so the assertion holds for the suite as written.
```

stevenspiel (approved):
LGTM

PR #320, opened 2026-09-08
refactor(SENTZ-5696): Close the certificate-pinning follow-ups from the PR #306 review

bcgreijnlautier:
Soundtrack of this PR: Talk Talk - Living In Another World

### Motivation

The PR #306 merge review left five follow-ups on the certificate-pinning code. `NetworkConfig` routes both trust-root setters through one `setTrustRoots` helper, which has a `WritableKeyPath`, a closure parameter and a cast that can't fail. An `SSLCertificateValidator` seam sits in front of that helper, and its one implementation forwards a single call.

### In this PR
* Each setter inlines the helper, so the setter parses the roots, pushes them to the requester, then stores them in the order the pinning contract needs.
* `SSLCertificateValidator`, `SecSSLCertificateValidator` and `TransportProtocol.certificateValidator` are deleted, which leaves `SecSSLCertificates.make` as the one route.
* `SSLCertificates.make` answers `Result<Self, InvalidInputError>`, and the extension's default `init?` is deleted. Each change breaks an external conformer at compile time, and the `Unreleased` section records both.
* Ten integration test factory sites take the requester from the config.
* A new unit case proves a refused fog push keeps the roots the call before it pinned.

Fixes https://mobilecoin-eng.atlassian.net/browse/SENTZ-5696

### Future Work
* https://mobilecoin-eng.atlassian.net/browse/SENTZ-5588 ships these `Unreleased` bullets in the dated 6.2.0 entry.

PR #322, opened 2026-09-09
fix(SENTZ-5683): Close the trust-root pinning follow-ups

bcgreijnlautier:
Soundtrack of this PR: [Lock and Key](https://open.spotify.com/search/lock%20and%20key)

### Motivation

The trust-root pinning work left four follow-ups open. The requester pins fog and consensus roots, and it doesn't have a mistyswap writer. It writes the three fields under three separate locks. It keeps empty host names and empty trust roots in a pinned set, and it logs whole certificates.

Fixes https://mobilecoin-eng.atlassian.net/browse/SENTZ-5683

### In this PR
* `setMistyswapTrustRoots` pins mistyswap roots, and `setAllTrustRoots` writes all three fields under one lock.
* An empty host name, an empty trust-root set and a hostless mistyswap pin are refused at the writer.
* The certificate log line has a byte count.
* `pinnedKeys` documents the merged-keys case, where a host that both fog and consensus name gets both sets.
* An IDN host shares a pin with its punycode form.
* `RefusingHttpRequester` moves beside its two call sites as a file-private mock, and the unused `certificateCount` goes.
* Every added test is proven sensitive by mutation, with the evidence on the ticket.

Stacked on #320, which reviews first.

### Future Work
* None.


Sources/Common/Network/HttpRequester.swift
```diff
@@ -23,4 +23,7 @@ public protocol HttpRequester {
     @discardableResult
     func setConsensusTrustRoots(_ trustRoots: SecSSLCertificates?, hosts: [String])
         -> Result<(), InvalidInputError>
+    @discardableResult
+    func setMistyswapTrustRoots(_ trustRoots: SecSSLCertificates?, hosts: [String])
+        -> Result<(), InvalidInputError>
```
stevenspiel, line 28: Implementers will need to add this function before it will compile, which is a breaking change. A line in the CHANGELOG.md may be in order.

Sources/HTTPS/DefaultHttpRequester.swift
```diff
@@ -103,23 +121,31 @@ extension DefaultHttpRequester {
...
     private let trustRoots = ReadWriteDispatchLock(TrustRoots())
 
     /// Answers with the keys of each set that names `host` and carries keys.
-    /// A host no such set names gets the keys of every set.
+    /// A host no such set names gets the keys of every set. A host both fog
+    /// and consensus name gets both sets' keys merged.
     func pinnedKeys(for host: String) -> [SecKey] {
         let name = CertificatePinningDelegate.normalized(host)
-        let roots = trustRoots.readSync { [$0.fog, $0.consensus] }
+        let roots = trustRoots.readSync { [$0.fog, $0.consensus, $0.mistyswap] }
```
stevenspiel, line 148: just noting that this and all other mistyswap references will get removed when mistyswap gets taken out. I can do that as a follow up once this is merged, or if you'd rather a cleaner git history, these changes can be reverted. Once the version [bump](https://github.com/mobilecoinofficial/libmobilecoin/pull/44) is merged and released, I'll have a PR ready.

stevenspiel (approved):
LGTM

Sources/Common/Network/HttpRequester.swift
bcgreijnlautier: `CHANGELOG.md` now records the third `HttpRequester` setter, the removed public `SecTrust.certificateCount` and the empty trust-root array that every `Config` setter refuses. Recording the setter also corrected three `[Unreleased]` lines from PR #320 that counted two.

Sources/HTTPS/DefaultHttpRequester.swift
bcgreijnlautier: I'll keep the mistyswap pinning and take your follow-up. Those hosts stay unpinned until the removal lands, and mobilecoinofficial/libmobilecoin#44 is still open, so a revert would trade a live gap for a tidier history.

